Issue #275, opened 2021-12-27
proposal: move jemalloc to optional feature


skyzh:
We should use a feature to enable jemalloc, instead of making it default.

rapiz1:
jemallocator doesn't support risc-v. Marking it as optional maybe facilitate compiling the project on risc-v 😯

likg227:
> jemallocator doesn't support risc-v. Marking it as optional maybe facilitate compiling the project on risc-v 😯

In fact, it also can't be compiled on my Mac (

PR #335, opened 2022-01-19
feat: separate jemalloc as a feature

rapiz1:
Separate jemalloc as a default feature

Resolves https://github.com/risinglightdb/risinglight/issues/275


skyzh (approved):
Thanks!

PR #337, opened 2022-01-19
ci: Run test on every push and PR

Xuanwo:


skyzh:
If someone creates a PR from inner repo, the workflow may run twice. The current configuration already supports run CI for every commit after a PR is created.

Xuanwo:
> If someone creates a PR from inner repo, the workflow may run twice. The current configuration already supports run CI for every commit after a PR is created.

I see. But I think they are different, one run for push and another one is for pull request.

PR #339, opened 2022-01-19
refactor(datachunk): replace prettytable-rs with comfy-table

PsiACE:
comfy-table is still actively maintained.

Signed-off-by: Chojan Shang <psiace@outlook.com>

### DIFF
--- a/src/array/data_chunk.rs
+++ b/src/array/data_chunk.rs
@@ -92,11 +92,11 @@ pub type DataChunkRef = Arc<DataChunk>;
 /// Print the chunk as a pretty table.
 impl fmt::Display for DataChunk {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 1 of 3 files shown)


src/array/data_chunk.rs
```diff
@@ -92,11 +92,11 @@ pub type DataChunkRef = Arc<DataChunk>;
 /// Print the chunk as a pretty table.
 impl fmt::Display for DataChunk {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        use prettytable::{format, Table};
+        use comfy_table::Table;
         let mut table = Table::new();
-        table.set_format(*format::consts::FORMAT_NO_LINESEP_WITH_TITLE);
+        table.load_preset("||--+-++|    ++++++");
```
PsiACE, line 97: The two should be equivalent.

skyzh (approved):
Thanks!

PR #342, opened 2022-01-19
feat(storage): support distinct value statistics

### LINKED ISSUE #254: storage: support distinct value statistics
RisingLight supports per-block statistics. Now we have a prototype for statistics framework, and we need to support more statistics.

* Implementing DistinctValue statistics https://github.com/singularity-data/risinglight/blob/main/src/storage/secondary/statistics/distinct_value.rs
* Build statistics in IndexBuilder https://github.com/singularity-data/risinglight/blob/main/src/storage/secondary/index_builder.rs
* Expose DistinctValue interface to user https://github.com/singularity-data/risinglight/blob/d8a6d5911a465810002a91ca23db168b9ce76d50/src/storage/secondary/transaction.rs#L267

ludics:
storage: support distinct value statistics #254

* add `distinct_count()` method for `BlockBuilder` trait
* update `PrimitiveColumnBuilder` and `CharColumnBuilder` to add `distinct_count`
* update `StatisticsGlobalAgg`

Signed-off-by: Lu Di <leonludics@gmail.com>

close #254


src/storage/secondary/block/primitive_nullable_block_builder.rs
```diff
@@ -52,6 +53,14 @@ impl<T: PrimitiveFixedWidthEncode> BlockBuilder<T::ArrayType>
         !self.data.is_empty() && self.estimated_size() + 1 + T::WIDTH > self.target_size
     }
 
+    fn distinct_count(&self) -> usize {
+        let mut distinct_values = HashSet::<&[u8]>::new();
+        for item in self.data.chunks(T::WIDTH) {
+            distinct_values.insert(item);
```
skyzh, line 59: This seems to ignore `None` value? Maybe we should also take the bitmap into account.

src/storage/secondary/statistics/distinct_value.rs
```diff
@@ -1,23 +1,36 @@
...
     }
 }
 
 impl StatisticsGlobalAgg for DistinctValueGlobalAgg {
-    fn apply_batch(&mut self, _index: &ColumnIndex) {}
+    fn apply_batch(&mut self, index: &ColumnIndex) {
+        for index in index.indexes() {
+            for stat in &index.stats {
+                if stat.block_stat_type() == BlockStatisticsType::DistinctValue {
+                    let cnt = usize::from_le_bytes(stat.body.clone().try_into().unwrap());
```
skyzh, line 25: I would suggest using `u64` instead of `usize` across all serialization places.

src/storage/secondary/block.rs
```diff
@@ -51,6 +51,9 @@ pub trait BlockBuilder<A: Array> {
     /// Get estimated size of block. Will be useful on runlength or compression encoding.
     fn estimated_size(&self) -> usize;
 
+    /// Get distinct values count of block
+    fn distinct_count(&self) -> usize;
```
skyzh, line 55: We will have `null_count` and many more statistics in the future. Can we squash all of them into one function?

e.g. `get_statistics(&self) -> Vec<BlockStatistics>`

skyzh (commented):
Rest LGTM, good work!

skyzh:
... by the way, the previous `StatisticsPartialAgg` seems not to be used any more? I think the current solution is much better than the `StatisticsPartialAgg` one, and would you please help remove those deprecated code?

src/storage/secondary/block/block_index_builder.rs
```diff
@@ -41,15 +42,20 @@ impl BlockIndexBuilder {
         block_type: BlockType,
         column_data: &mut Vec<u8>,
         block_data: &mut Vec<u8>,
+        distinct_count: u64,
```
skyzh, line 45: Should also take `Vec<BlockStatistics>` as input parameter?

skyzh (approved):
LGTM, good work!

* Would you please test if `\stat table column` can correctly print the distinct value?
* It seems that every block iterator has its own logic of get statistics. We can implement something like:

```
pub struct StatisticsBuilder<'a> {
  distinct_values: HashSet::<&[u8]>
}

impl <'a> StatisticsBuilder <'a> {
  pub fn add_item(&mut self, data: Option<&[u8]>) {
    ...
  }

  pub fn get_statistics(self) -> Vec<BlockStatistics> {
    ...
  }
}
```

... which can be used in every block builder and reduce duplicated code.

The current PR already looks good to me, you may get it merged as you want. (Do you have the permission to merge? If not, ping me to grant you write access.) After that, you may reduce duplicated code by adding `StatisticsBuilder` in later PRs.

ludics:
> LGTM, good work!
> 
> * Would you please test if `\stat table column` can correctly print the distinct value?
> * It seems that every block iterator has its own logic of get statistics. We can implement something like:
> 
> ```rust
> pub struct StatisticsBuilder<'a> {
>   distinct_values: HashSet::<&'a [u8]>
> }
> 
> impl <'a> StatisticsBuilder <'a> {
>   pub fn add_item(&mut self, data: Option<&[u8]>) {
>     ...
>   }
> 
>   pub fn get_statistics(self) -> Vec<BlockStatistics> {
>     ...
>   }
> }
> ```
> 
> ... which can be used in every block builder and reduce duplicated code.
> 
> The current PR already looks good to me, you may get it merged as you want. (Do you have the permission to merge? If not, ping me to grant you write access.) After that, you may reduce duplicated code by adding `StatisticsBuilder` in later PRs.

* I've tested `\stat table column`, data types including int, char, varchar, bool, decimal, and the outputs within my testing range are as expected
* I don't have permission to merge 😩

ludics:
Implemented StatisticsBuilder, and would you like to review it?

skyzh (approved):
LGTM (again!)

PR #353, opened 2022-01-23
feat(join): refactor and support outer join for hash join

wangrunji0408:
This PR refactors `NestedLoopJoin` and `HashJoin` executors, and adds outer join support to hash join.

### DIFF
--- a/src/executor/hash_join.rs
+++ b/src/executor/hash_join.rs
@@ -3,8 +3,10 @@
 use std::collections::HashMap;
 use std::vec::Vec;
 
+use futures::TryStreamExt;
+
 use super::*;
-use crate::array::{ArrayBuilderImpl, DataChunk};
+use crate::array::{ArrayBuilderImpl, DataChunk, RowRef};
 use crate::binder::{BoundExpr, BoundJoinOperator};
 use crate::types::{DataType, DataValue};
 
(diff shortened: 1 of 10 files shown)


src/executor/nested_loop_join.rs
```diff
@@ -21,170 +21,94 @@ pub struct NestedLoopJoinExecutor {
...
+            ArrayImpl::Bool(a) => a,
+            _ => panic!("unsupported value from join condition"),
         };
+        yield cross_chunk.filter(filter.iter().map(|b| matches!(b, Some(true))));
 
-        for left_chunk_idx in 0..left_chunks.len() {
-            for left_row_idx in 0..left_chunks[left_chunk_idx].cardinality() {
+        // append rows for left outer join
+        if matches!(
+            self.join_op,
```
TennyZhuang, line 62: I prefer treat join_op as const generics, but we can refine it later.

src/executor/nested_loop_join.rs
```diff
@@ -21,170 +21,94 @@ pub struct NestedLoopJoinExecutor {
...
+            BoundJoinOperator::RightOuter | BoundJoinOperator::FullOuter
+        ) {
+            let mut builders = create_builders();
+            // scan the filter to find whether a right row matches any left rows
+            let mut matched = bitvec![0; right_rows().count()];
+            let mut i = 0;
+            for _ in left_rows() {
+                for (j, _) in right_rows().enumerate() {
```
TennyZhuang, line 96: Why not 0..right_rows().len()?

src/executor/nested_loop_join.rs
```diff
@@ -21,170 +21,94 @@ pub struct NestedLoopJoinExecutor {
...
+                    if matches!(filter.get(i), Some(true)) {
+                        matched.set(j, true);
                     }
+                    i += 1;
```
TennyZhuang, line 100: Why maintain I manually?

src/executor/nested_loop_join.rs
wangrunji0408: Good idea...

src/executor/nested_loop_join.rs
wangrunji0408: I need something like `.enumerate()` in the nested for-loop. This way is the simplest I can think (

TennyZhuang: got it

TennyZhuang: How about https://docs.rs/itertools/0.4.8/itertools/trait.Itertools.html#method.cartesian_product?

TennyZhuang: catesian_product().enumerate(), then we may got i?

Not necessary, any one is ok

src/executor/nested_loop_join.rs
wangrunji0408: Sounds good. But if we do that, we have to construct 4 types when building the executor...

TennyZhuang: Yes, @yuhao-su has done it in another repo.

TennyZhuang: We can do it after adt_const_generics completed.

src/executor/nested_loop_join.rs
wangrunji0408: I tried:
```rust
for (i, (_, (j, _))) in left_rows()
    .cartesian_product(right_rows().enumerate())
    .enumerate()
{
    if matches!(filter.get(i), Some(true)) {
        matched.set(j, true);
    }
}
```

However, `.cartesian_product(other)` requires `other` to be cloneable, unfortunately that is impossible in our case.

Give up 🤪

wangrunji0408: I'm stupid. It works now:
```rust
for (i, (_, j)) in left_rows()
    .cartesian_product(0..right_rows().count())
    .enumerate()
{
    if matches!(filter.get(i), Some(true)) {
        matched.set(j, true);
    }
}
```

wangrunji0408: I found another way which can avoid bitmap: a46215c

src/executor/hash_join.rs
```diff
@@ -13,89 +15,91 @@ pub struct HashJoinExecutor {
...
-            chunk_builders.push(ArrayBuilderImpl::new(ty));
+
+        // probe
+        let mut builders = (self.left_types.iter())
+            .chain(self.right_types.iter())
+            .map(|ty| ArrayBuilderImpl::with_capacity(PROCESSING_WINDOW_SIZE, ty))
+            .collect_vec();
+        for right_row in right_rows() {
+            let hash_value = right_row.get(self.right_column_index);
+            for left_row in hash_map.get(&hash_value).unwrap_or(&vec![]) {
```
skyzh, line 54: I would prefer `unwrap_or_else` or create the empty vector once at the beginning of this function.

src/executor/hash_join.rs
```diff
@@ -13,89 +15,91 @@ pub struct HashJoinExecutor {
...
+                let values = left_row.values().chain(right_row.values());
+                for (builder, v) in builders.iter_mut().zip(values) {
```
skyzh, line 56: consider using `Itertools::zip_eq`.

skyzh (approved):
Rest LGTM, good work!

src/executor/hash_join.rs
wangrunji0408: `vec![]` (`Vec::new()`) is a const fn. So I guess the compiler should be able to optimize here?

skyzh: Get it! Then I think there's no need to optimize :)

TennyZhuang (approved):
LGTM

PR #357, opened 2022-01-24
feat(optimizer): activate rules

MingjiHan99:
Signed-off-by: MingjiHan <mingjihan@umass.edu>

### DIFF
--- a/src/optimizer/logical_plan_rewriter/arith_expr_simplification.rs
+++ b/src/optimizer/logical_plan_rewriter/arith_expr_simplification.rs
@@ -14,10 +14,10 @@ use crate::types::DataValue::*;
 /// `select 1 * a, b / 1, c + 0, d - 0 from t;`
 /// The query will be converted to:
 /// `select a, b, c, d from t;`
-pub struct ArithExprSimplification;
+pub struct ArithExprSimplificationRule;
 
-impl Rewriter for ArithExprSimplification {
-    fn rewrite_expr(&mut self, expr: &mut BoundExpr) {
+impl ExprRewriter for ArithExprSimplificationRule {
+    fn rewrite_expr(&self, expr: &mut BoundExpr) {
         // TODO: support more data types.
         let new = match &expr {
             BinaryOp(op) => match (&op.op, &*op.left_expr, &*op.right_expr) {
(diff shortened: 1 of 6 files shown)


src/optimizer/mod.rs
```diff
@@ -32,7 +32,14 @@ impl Optimizer {
         // plan = plan.rewrite(&mut ConstantFolding);
         // plan = plan.rewrite(&mut ArithExprSimplification);
         // plan = plan.rewrite(&mut BoolExprSimplification);
```
st1page, line 34:
```suggestion
```

skyzh (approved):
LGTM. Would you please fix fmt and (if possible) add test cases for the rewriters?

src/optimizer/logical_plan_rewriter/arith_expr_simplification.rs
```diff
@@ -62,3 +62,34 @@ impl Rewriter for ArithExprSimplification {
         *expr = new;
     }
 }
+
+impl PlanRewriter for ArithExprSimplificationRule {
+    fn rewrite_logical_join(&mut self, join: &LogicalJoin) -> PlanRef {
+        let left = self.rewrite(join.left());
+        let right = self.rewrite(join.right());
+        Arc::new(join.clone_with_rewrite_expr(left, right, self))
+    }
```
wangrunji0408, line 71: I don't think it's elegant to manually implement these functions for pure expression rules.
Is there a way to resolve that? cc @st1page

st1page: maybe we could implement a kind of `PlanRewriter`, give it a Box<ExprRewriter> when initalizing it. all Expression rewrite demand will use the same code.

st1page: or another way, how about doing expression simplificatation when construct the expression? so we can make sure any expression appearing in our project has been simplificated. just a half-baked idea.

st1page: Anyway, we can refactor in future PRs.

wangrunji0408: > or another way, how about doing expression simplificatation when construct the expression? so we can make sure any expression appearing in our project has been simplificated. just a half-baked idea.

This way we will lose the original expressions, not good for debugging. -1 (

MingjiHan99:
> LGTM. Would you please fix fmt and (if possible) add test cases for the rewriters?

We will add them in the incoming PR.

PR #359, opened 2022-01-25
feat(binder): insert type cast on implicit type conversion

### LINKED ISSUE #241: binder: remove Decimal conversion
Currently, when users do insert int into a decimal column by:

```
insert into table values (1)
```

where table has a single decimal column, we will convert users' value into Decimal in binder. This is okay for now. But if we want to support insert into ... select, we have to insert a cast to plan. Therefore, to unify both cases, we should not do any cast in binder.

pleiadesian:
close #241


skyzh (approved):
Never thought that the final solution can be this simple 🤣

PR #363, opened 2022-01-26
ci: use self-hosted runners

skyzh:


skyzh:
... maintainers should notice that now every PR from first contributors should be approved by ourselves.

skyzh:
Found several bugs with the current approach. The terraform module doesn't fit our need very well. I would disable it.

PR #369, opened 2022-01-27
feat(optimizer): add optimizer test framework

MingjiHan99:
Signed-off-by: MingjiHan <mingjihan@umass.edu>

### DIFF
--- a/src/optimizer/logical_plan_rewriter/arith_expr_simplification.rs
+++ b/src/optimizer/logical_plan_rewriter/arith_expr_simplification.rs
@@ -91,3 +91,76 @@ impl PlanRewriter for ArithExprSimplificationRule {
         Arc::new(plan.clone_with_rewrite_expr(self))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use serde::Serialize;
+
+    use crate::Database;
+    #[tokio::main]
+    #[test]
+    async fn test_expr_simplification() {
+        let db = Database::new_in_memory();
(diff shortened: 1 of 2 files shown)


wangrunji0408 (approved):
LGTM

src/optimizer/logical_plan_rewriter/arith_expr_simplification.rs
```diff
@@ -91,3 +91,76 @@ impl PlanRewriter for ArithExprSimplificationRule {
...
+
+        let plans = db.generate_execution_plan(sql0).unwrap();
+        assert_eq!(plans.len(), 1);
+        let buf = Vec::new();
+        let formatter = serde_json::ser::PrettyFormatter::with_indent(b"    ");
+        let mut ser = serde_json::Serializer::with_formatter(buf, formatter);
+        plans[0].serialize(&mut ser).unwrap();
+        let ser_str = String::from_utf8(ser.into_inner()).unwrap();
+        println!("{}", ser_str);
+        assert_eq!(
```
skyzh, line 117: I don't quite like writing so much json inside our code. I would suggest either:

* Make it into a file. e.g, `assert_eq!(serde_json::deserialize(include_bytes!("xxx.json")), json_generated_in_test)`
* Use `json!` macro.

We should also provide ways for future developers to update all plans at once. I think asserting plan eq is simply a regression test, so developers should be able to update all and diff to see if everything is as their expectation.

skyzh (approved):
Rest LGTM, good work!

Issue #366, opened 2022-01-27
executor: support multi-thread execution


skyzh:
For the first step, we don't do data shuffle -- we simply distribute each executor on different threads.

PR #367, opened 2022-01-27
feat(optimizer): add exchange plan

skyzh:
Signed-off-by: Alex Chi <iskyzh@gmail.com>

part of https://github.com/risinglightdb/risinglight/issues/366

As the optimizer will visit children recursively, we have to set a `exchanged` flag on `LogicalTableScan`, so that it won't be exchanged further.

Not sure if this is the correct way to do this. Any way to stop iteration (or finish the rule)?

PR #371, opened 2022-01-28
feat(storage): support multi-thread execution for table scan

skyzh:

part of https://github.com/risinglightdb/risinglight/issues/366

### DIFF
--- a/src/executor/table_scan.rs
+++ b/src/executor/table_scan.rs
@@ -53,7 +53,7 @@ impl<S: Storage> TableScanExecutor<S> {
     }
 
     #[try_stream(boxed, ok = DataChunk, error = ExecutorError)]
-    pub async fn execute(self) {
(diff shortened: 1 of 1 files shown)


TennyZhuang (approved):
LGTM

wangrunji0408 (commented):
We can do this in `ExecutorBuilder` so that all executors will benefit.

skyzh:
> We can do this in `ExecutorBuilder` so that all executors will benefit.

For the more general case, I think we should use optimizer to decide which node to exchange with a `PhysicalExchange` plan node. See https://github.com/risinglightdb/risinglight/pull/367.

st1page:
> We can do this in `ExecutorBuilder` so that all executors will benefit.

https://github.com/risinglightdb/risinglight/pull/367#pullrequestreview-866848198

PR #373, opened 2022-01-28
feat: support TPC-H Q5 & Q10

wangrunji0408:
This PR adds TPC-H Q5 and fixes `date + interval (month)` to pass TPC-H Q10.

### DIFF
--- a/tests/sqllogictest.rs
+++ b/tests/sqllogictest.rs
@@ -35,7 +35,10 @@ fn test_mem(name: &str) {
         db: Database::new_in_memory(),
     });
     tester.enable_testdir();
-    tester.run_file(Path::new("tests/sql").join(name)).unwrap();
+    tester
+        .run_file(Path::new("tests/sql").join(name))
+        .map_err(|e| panic!("{}", e))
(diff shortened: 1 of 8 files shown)


tests/sqllogictest.rs
```diff
@@ -67,7 +70,10 @@ fn test_disk(name: &str) {
     ));
     let mut tester = sqllogictest::Runner::new(DatabaseWrapper { rt, db });
     tester.enable_testdir();
-    tester.run_file(Path::new("tests/sql").join(name)).unwrap();
+    tester
+        .run_file(Path::new("tests/sql").join(name))
+        .map_err(|e| panic!("{}", e))
```
skyzh, line 75: ... why panic before unwrap?

skyzh (approved):
Rest LGTM, good work!

pleiadesian (approved):

tests/sqllogictest.rs
wangrunji0408: To print the error in Display format, which is more readable than Debug format...

Issue #350, opened 2022-01-23
test: enable `select.slt` test


skyzh:
I didn't even remember why we disabled it before 🤣🤣🤣

skyzh:
Hmm... seems not resolved?

PR #376, opened 2022-01-28
test: auto discover test cases from path

wangrunji0408:
This PR enables the rest of `basic_test.slt` and a part of `select.slt` (#350) tests.

This also resolves #338: automatically discover test cases from path, by generating source file in build.rs.
Note that there is a test `statistics.slt` that fails on in-memory storage, it has been added to the blocklist in build.rs.


skyzh (approved):
LGTM, good work! 

* Should also update and test docs/tpch.md. In that, we used `create.slt`. As `.slt` now has different suffix, the rising light binary will regard it as SQL. Need some changes.
* Should update `contributing.md` so that new developers know that `.slt` will be automatically picked up.

wangrunji0408:
I have changed the ignore pattern from `*.slt.0` to `_*.slt`. 
Please take a look @skyzh

skyzh (approved):

PR #386, opened 2022-01-30
feat: add initial support for blob type

wangrunji0408:
This PR adds initial support for blob type.

We borrowed a test script from DuckDB and managed to pass most of the tests on memory storage.
The failed case is `SELECT NULL::BLOB`, which needs further fixes in our binder.

This PR also proposes another approach to support unsized types in array without using GAT (related discussion #381).
In short, we can define unsized types over a slice, like `std::path::Path`. They have the same memory layout as `[u8]` but can parse the underlying data in their own way. For more details, you can see `BlobRef` in the code.

TODO: support blob type on disk storage.

### DIFF
--- a/src/array/utf8_array.rs
+++ b/src/array/utf8_array.rs
@@ -1,28 +1,62 @@
 // Copyright 2022 RisingLight Project Authors. Licensed under Apache-2.0.
 
 use std::iter::FromIterator;
+use std::marker::PhantomData;
 
 use bitvec::vec::BitVec;
 use serde::{Deserialize, Serialize};
 
 use super::{Array, ArrayBuilder, ArrayEstimateExt, ArrayValidExt};
 
-/// A collection of Rust UTF8 `String`s.
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub struct Utf8Array {
+/// A collection of variable-length values.
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub struct BytesArray<T: ValueRef + ?Sized> {
     offset: Vec<usize>,
     valid: BitVec,
     data: Vec<u8>,
+    _type: PhantomData<T>,
 }
 
-impl Array for Utf8Array {
-    type Item = str;
-    type Builder = Utf8ArrayBuilder;
+/// The borrowed type of a variable-length value.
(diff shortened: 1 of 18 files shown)


skyzh (approved):
LGTM! Looks like a good way that we can get rid of the expensive type exercise 🤪🤪🤪

src/array/utf8_array.rs
```diff
@@ -1,28 +1,64 @@
...
+    fn from_bytes(s: &[u8]) -> &Self;
+}
+
+impl ValueRef for str {
+    fn from_bytes(s: &[u8]) -> &Self {
+        unsafe { std::str::from_utf8_unchecked(s) }
+    }
+}
+
+impl ValueRef for BlobRef {
```
skyzh, line 32: Can we implement `ValueRef` directly on `[u8]`, so that we can get rid of the unsafe transmute between `BlobRef` and `&[u8]`?

skyzh: Per discussion at #381, what if in the future we want to store something like `blob_type` in `Blob`? If we define `BlobRef` as:

```
pub struct BlobRef(BlobType, [u8]);
```

Can we still do the transmute like this?

wangrunji0408: > Can we implement `ValueRef` directly on `[u8]`, so that we can get rid of the unsafe transmute between `BlobRef` and `&[u8]`?

Yes, but this way has some limitations, such as `&[u8]` doesn't implement `Display` which will bring trouble to compilation. We need to customize the output format on the new type. And the unsafe transmute from `&[u8]` to `&T: ValueRef` is unavoidable, because not all bytes are valid `T`. Only the bytes in array can be safely converted.

> Per discussion at #381, what if in the future we want to store something like blob_type in Blob? If we define BlobRef as:
> ```
> pub struct BlobRef(BlobType, [u8]);
> ```
> Can we still do the transmute like this?

Nope, the new type must be defined exactly as `[u8]`, in order to have the same memory layout. This is the limitation of this approach. 
If we need something like `blob_type`, we should encode it into the `[u8]`. For example we can encode the first byte as type and the rest as content, then the code will be:
```rust
pub struct BlobRef([u8]);

impl BlobRef {
    pub fn type_(&self) -> u8 {
        self.0[0]
    }
    pub fn content(&self) -> &[u8] {
        &self.0[1..]
    }
}
```

Issue #389, opened 2022-01-31
storage: remove `async_trait` and manually expand return-position `Future`


skyzh:
So that we can have lower overhead when calling `.next_batch`!

You can find some useful snippets from [skyzh's blog](https://www.skyzh.dev/posts/articles/2022-01-31-gat-async-trait/)

- [x] #392
- [x] #393
- [x] #395
- [x] #394
- [x] `TxnIterator` trait https://github.com/risinglightdb/risinglight/pull/391

TennyZhuang:
I'd like to try it.

PR #391, opened 2022-02-01
refactor(storage): use static async trait in TxnIterator

TennyZhuang:
Part of #389 

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -12,6 +12,7 @@ mod error;
 pub use error::{StorageError, StorageResult, TracedStorageError};
 
 mod chunk;
+use std::future::Future;
 use std::sync::Arc;
 
 use async_trait::async_trait;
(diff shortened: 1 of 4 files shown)


src/storage/mod.rs
```diff
@@ -147,11 +148,11 @@ pub trait Transaction: Sync + Send + 'static {
 }
 
 /// An iterator over table in a transaction.
-#[async_trait]
 pub trait TxnIterator: Send {
+    type NextFuture<'a>: Future<Output = StorageResult<Option<DataChunk>>> + Send
```
skyzh, line 152: Better to add a `'a` bound for `NextFuture`?

skyzh (approved):
Rest LGTM

src/storage/mod.rs
skyzh: We can for sure eliminate `'a` here, as `type NextFuture<'a> = impl Future<Output = StorageResult<Option<DataChunk>>>;` will automatically `+ 'a` when being put in the return position. Anyway, a `+ 'a` here can make reader less confusing about: "why there is an `'a` but not used?"

PR #397, opened 2022-02-01
deprecated async_trait for Transaction trait

### LINKED ISSUE #395: storage: remove `async_trait` for `Transaction` trait

RishiKumarRay:
Fixed #395

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -114,7 +114,6 @@ pub trait RowHandler: Sync + Send + 'static {
 }
 
 /// Represents a transaction in storage engine.
-#[async_trait]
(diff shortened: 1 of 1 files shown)


TennyZhuang:
> Fixed #395

The issue means that we should use a static strategy to dispatch the async method, not just remove the proc-macro. It will cause the compilation failed.

A good example can be found at #391 

Please make sure your codes can be compiled and the tests passed correctly before request a review, thanks!

You can find [contribution guide](https://github.com/risinglightdb/risinglight/blob/main/CONTRIBUTING.md) here.

TennyZhuang:
The issue has been done correctly in #405

PR #401, opened 2022-02-02
refactor(storage): use static async trait in ColumnIterator

### LINKED ISSUE #394: storage: remove `async_trait` for `ColumnIterator` trait

zzl200012:
Fixed #394 

### DIFF
--- a/src/storage/secondary/column.rs
+++ b/src/storage/secondary/column.rs
@@ -75,7 +75,7 @@ pub trait ColumnIterator<A: Array> {
     /// iterating, the returned value should be 0.
     fn fetch_hint(&self) -> usize;
 
-    /// Skip 'cnt' items for this column_iterator.
+    /// Skip 'cnt' items for this `column_iterator`.
     fn skip(&mut self, cnt: usize);
 }
 
--- a/src/storage/secondary/column.rs
+++ b/src/storage/secondary/column.rs
@@ -15,8 +15,8 @@ mod primitive_column_builder;
 mod primitive_column_factory;
 mod row_handler_sequencer;
 
-use std::io::{Read, Seek, SeekFrom};
 use std::future::Future;
+use std::io::{Read, Seek, SeekFrom};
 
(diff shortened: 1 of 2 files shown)


src/storage/secondary/column.rs
```diff
@@ -57,22 +57,25 @@ pub trait ColumnBuilder<A: Array> {
...
+    type NextFuture<'a>: Future<Output = StorageResult<Option<(u32, A)>>> + 'a
+    where
+        Self: 'a;
+
     /// Get a batch and the starting row id from the column. A `None` return value means that
     /// there are no more elements from the block. By using `expected_size`, developers can
     /// get an array of NO MORE THAN the `expected_size` on supported column types.
-    async fn next_batch(
-        &mut self,
+    fn next_batch<'a>(
```
skyzh, line 68:
```suggestion
    fn next_batch(
```

No need to explicitly write `'a` here.

In the below parameters, simply use `&mut self`, `Option<&BitVec>`, and `Self::NextFuture<'_>`.

skyzh (commented):
rest LGTM

src/storage/secondary/column.rs
zzl200012: Sorry, I'm a beginner of rust. I tried this suggestion, but compiler said:
```
error[E0623]: lifetime mismatch
   --> src/storage/secondary/column/concrete_column_iterator.rs:298:10
    |
297 |         filter_bitmap: Option<&BitVec>,
    |                               ------- this parameter and the return type are declared with different lifetimes...
298 |     ) -> Self::NextFuture<'_> {
    |          ^^^^^^^^^^^^^^^^^^^^
    |          |
    |          ...but data from `filter_bitmap` is returned here
```
Is there any way to fix that?

skyzh: hmmm... Looks like a compiler bug. No need to fix.

skyzh (approved):

Issue #393, opened 2022-02-01
storage: remove `async_trait` for `Table` trait


TennyZhuang:

tabVersion:
I'd like to have a try.

PR #403, opened 2022-02-03
refactor(table): remove async_trait for Table trait

tabVersion:
part of https://github.com/risinglightdb/risinglight/issues/389

Issue Number: close https://github.com/risinglightdb/risinglight/issues/393

Signed-off-by: tabVersion tabvision@bupt.icu

--- 

one question: 

Since `ReadResultFuture`, `WriteResultFuture` and `UpdateResultFuture` share type `Future<Output = StorageResult<Self::TransactionType>>`, why rust requires implementing them separately.

### DIFF
--- a/src/storage/memory/table.rs
+++ b/src/storage/memory/table.rs
@@ -4,7 +4,7 @@ use std::collections::HashSet;
 use std::sync::{Arc, RwLock};
 use std::vec::Vec;
 
-use async_trait::async_trait;
+use futures::Future;
 
 use super::*;
 use crate::array::DataChunk;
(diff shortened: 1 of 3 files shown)


src/storage/memory/table.rs
```diff
@@ -64,9 +64,14 @@ impl InMemoryTable {
     }
 }
 
-#[async_trait]
 impl Table for InMemoryTable {
     type TransactionType = InMemoryTransaction;
+    type ReadResultFuture<'a> =
+        impl Future<Output = StorageResult<Self::TransactionType>> + Sync + Send + 'a;
```
skyzh, line 70: No need to `Sync`?

skyzh (commented):
Rest LGTM, good work!

skyzh:
> Since ReadResultFuture, WriteResultFuture and UpdateResultFuture share type Future<Output = StorageResult<Self::TransactionType>>, why rust requires implementing them separately.

They share the same trait `Future<Output = StorageResult<Self::TransactionType>>`, but the underlying type is different for each function.

src/storage/memory/table.rs
tabVersion: fixed

src/storage/mod.rs
```diff
@@ -85,22 +85,40 @@ pub trait Storage: Sync + Send + 'static {
...
+    where
+        Self: 'a;
+    type ReadResultFuture<'a>: Future<Output = StorageResult<Self::TransactionType>>
+        + Sync
+        + Send
+        + 'a
+    where
+        Self: 'a;
+    type UpdateResultFuture<'a>: Future<Output = StorageResult<Self::TransactionType>>
+        + Sync
```
skyzh, line 105: No need to `Sync`?

skyzh (commented):
Rest LGTM

src/storage/mod.rs
tabVersion: fixed

skyzh (approved):

Issue #406, opened 2022-02-03
docs: enable debug logs


Fedomn:
currently, the `export RUST_LOG=debug` command cannot enable risinglight debug logs.
so need to adjust the [environment variable in doc](https://github.com/risinglightdb/risinglight/blob/main/docs/00-develop.md?plain=1#L63)

PR #407, opened 2022-02-03
fix(docs): enable debug logs

Fedomn:
To close https://github.com/risinglightdb/risinglight/issues/406

### DIFF
--- a/docs/00-develop.md
+++ b/docs/00-develop.md
@@ -60,7 +60,7 @@ the official Rust extension. You will need to uninstall "The Rust Programming La
 To enable debug logs for RisingLight, export the following environment variable to your shell environment.
 
 ```
-export RUST_LOG=debug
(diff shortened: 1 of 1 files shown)


skyzh (commented):
We are using https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html for this config. It seems to be compatible with `RUST_LOG=debug`...

skyzh (approved):
I think this is caused by

```
.add_directive(Level::INFO.into());
```

After removing this statement, all debug logs, besides RisingLight's ones, are enabled by simply setting `RUST_LOG=debug`.

image

Considering there are too many rustyline's debug log, I would prefer to go with the current solution by simply setting `export RUST_LOG=risinglight=debug`, even if this seems like a hack :)

PR #411, opened 2022-02-05
feat(optimizer): use json file for optimizer testing

MingjiHan99:
Use json file for optimizer testing so we don't need to dump much json string in the codebase.

### DIFF
--- a/src/optimizer/logical_plan_rewriter/arith_expr_simplification.rs
+++ b/src/optimizer/logical_plan_rewriter/arith_expr_simplification.rs
@@ -94,9 +94,13 @@ impl PlanRewriter for ArithExprSimplificationRule {
 
 #[cfg(test)]
 mod tests {
+    use std::fs;
+    use std::path::{Path, PathBuf};
+
     use serde::Serialize;
 
     use crate::Database;
+
     #[tokio::main]
     #[test]
     async fn test_expr_simplification() {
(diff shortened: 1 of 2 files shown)


src/optimizer/logical_plan_rewriter/arith_expr_simplification.rs
```diff
@@ -113,54 +117,11 @@ mod tests {
...
-                \"with_row_handler\": false,
-                \"is_sorted\": false,
-                \"expr\": null
-            }
-        }
-    }
-}"
-        );
+        let mut d = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        d.push("tests/json/arith_expr_simplification.json");
```
skyzh, line 121: I would prefer using `include_str` macro to embed the file into the test binary in compile time.

skyzh (commented):
Would you please fix clippy?

TennyZhuang:
+1 for include_str!

skyzh (approved):
LGTM for the current patch. Anyway, I still recommend:

* using `include_str` macro in Rust std library. By using this way, we don't need to know where is the manifest file. It's relative to the current file.
* `serde_json` might not produce the same result for the same struct every time. The construction of hashmap, etc, would impact the result order. Consider convert the JSON file and the plan struct into `serde_json::Value` struct, which is comparable.
* If there's an update in plan struct, is there a convenient way to update all plan JSON in one click? When we are adding more and more plan tests, it would be hard to modify each JSON by hand.

Issue #398, opened 2022-02-01
bug: `create table t;` should directly throw exception


yingjunwu:
In standard SQL, users must provide table schema when creating a database. However, in RisingLight, we still allow the following command:
```
> create table t;
+---+
| 0 |
+---+
```
And it then directly leads to system crash:
```
> select * from t;
thread 'tokio-runtime-worker' panicked at 'assertion failed: !arrays.is_empty()', src/array/data_chunk.rs:21:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'failed to join scan thread: JoinError::Panic(...)', src/executor/table_scan.rs:129:23
```

We should throw exception if users do not provide schema. e.g., in pg:
```
mydb=> create table t;
2022-02-01 22:32:38.328 CST [6582] ERROR:  syntax error at or near ";" at character 15
2022-02-01 22:32:38.328 CST [6582] STATEMENT:  create table t;
ERROR:  syntax error at or near ";"
LINE 1: create table t;
                      ^
```

PR #413, opened 2022-02-06
fix(array): support no columns table

Fedomn:
Try to close https://github.com/risinglightdb/risinglight/issues/398

This table creation syntax is not standard and weird but supported by PostgreSQL. like this:

```
postgres@127:db1> create table t();
CREATE TABLE
Time: 0.006s

postgres@127:db1> select * from t;
SELECT 0
Time: 0.002s
```

After investigation, I found sqlparser-rs also supported this syntax, see https://github.com/sqlparser-rs/sqlparser-rs/pull/94. so RisingLight could parse successfully.

I think RisingLight should support the query `no columns table` instead of throwing exception on table creation.

So I just fixed the code that caused panic to avoid process termination.


TennyZhuang (approved):
Thanks for your contribution!

I've modified a few codes to make it more rusty, and push them directly.

LGTM

PR #414, opened 2022-02-06
refactor(storage): build rowset while appending data in ColumnMemTable

### LINKED ISSUE #409: storage: build rowset instead of buffering data when appending chunks
Currently, when building an in-memory `ColumnMemTable` rowset,

https://github.com/risinglightdb/risinglight/blob/78caff0ef7937a0b86200cf74c6672eff6e197ee/src/storage/secondary/rowset/mem_rowset.rs#L88-L90

We simply use in-memory array builder to store information, and...

https://github.com/risinglightdb/risinglight/blob/78caff0ef7937a0b86200cf74c6672eff6e197ee/src/storage/secondary/rowset/mem_rowset.rs#L132-L144

We encode them all-at-once at the end.

This would make the ingestion process not-so-efficient. We should be able to encode the rowset when appending chunks into the `ColumnMemTable`, instead of all-at-once.

zzl200012:

Fixed #409

### DIFF
--- a/src/storage/secondary/transaction.rs
+++ b/src/storage/secondary/transaction.rs
@@ -97,17 +97,11 @@ impl SecondaryTransaction {
         } else {
             return Ok(());
         };
-        let rowset_id = self.table.generate_rowset_id();
+        let rowset_id = self.table.get_current_rowset_id();
         let directory = self.table.get_rowset_path(rowset_id);
 
-        tokio::fs::create_dir(&directory).await?;
-
         // flush data to disk
-        mem.flush(
-            &directory,
-            ColumnBuilderOptions::from_storage_options(&*self.table.storage_options),
-        )
-        .await?;
+        mem.flush().await?;
 
         let on_disk = DiskRowset::open(
             directory,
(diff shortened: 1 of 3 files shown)


TennyZhuang:
Please format your codes with “cargo fly”

src/storage/secondary/transaction.rs
```diff
@@ -97,17 +97,11 @@ impl SecondaryTransaction {
         } else {
             return Ok(());
         };
-        let rowset_id = self.table.generate_rowset_id();
+        let rowset_id = self.table.get_current_rowset_id();
```
skyzh, line 100: This would cause problems in multi-thread cases. Better to store `rowset_id` in the memtable struct.

skyzh (commented):
Rest LGTM, good work!

src/storage/secondary/transaction.rs
zzl200012: updated

skyzh (approved):
LGTM, thanks!

Issue #388, opened 2022-01-31
expression: do not bind column for `count(*)` in binder


skyzh:
https://github.com/risinglightdb/risinglight/blob/40db687e7424da4ef96661c9cfe65adcbcf5a1a6/src/binder/expression/agg_call.rs#L65-L83

Currently, we always select the 0th column in binder for `count(*)`. This has caused problems like https://github.com/risinglightdb/risinglight/issues/200. In the future, we should be able to do both of the following:

* bind all columns in binder, and do column prune in optimizer
* rewrite `count(*)` to using statistics information (only for queries like `select count(*) from table`)

likg227:
I will solve this problem after @st1page finishes his work.

PR #420, opened 2022-02-07
feat(optimizer): column pruning framework

st1page:
add `prune_col` method on `PlanNode`, every logical node should implement `fn prune_col(&self, required_cols: BitSet) -> PlanRef`, which will optimize itself and output only the required columns. Now there is a default implementation which is insert a `LogialProjection` on the top of the Node.

ref https://github.com/risinglightdb/risinglight/issues/388

### DIFF
--- a/src/optimizer/mod.rs
+++ b/src/optimizer/mod.rs
@@ -1,5 +1,7 @@
 // Copyright 2022 RisingLight Project Authors. Licensed under Apache-2.0.
 
+use bit_set::BitSet;
+
 use crate::binder::*;
 
 mod expr_utils;
@@ -42,6 +44,8 @@ impl Optimizer {
         }
         let hep_optimizer = HeuristicOptimizer { rules };
         plan = hep_optimizer.optimize(plan);
+        let out_types_num = plan.out_types().len();
+        plan = plan.prune_col(BitSet::from_iter((0..out_types_num).into_iter()));
         let mut phy_converter = PhysicalConverter;
         phy_converter.rewrite(plan)
     }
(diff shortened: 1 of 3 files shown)


src/optimizer/mod.rs
```diff
@@ -42,6 +44,8 @@ impl Optimizer {
         }
         let hep_optimizer = HeuristicOptimizer { rules };
         plan = hep_optimizer.optimize(plan);
+        let out_types_num = plan.out_types().len();
+        plan = plan.prune_col(BitSet::from_iter(0..out_types_num));
```
skyzh, line 48: I think we already have `bitvec`, and there's no need to import another bit-vec-like crate...

src/optimizer/plan_nodes/mod.rs
```diff
@@ -87,13 +88,50 @@ pub use physical_simple_agg::*;
...
+        let exprs = required_cols
+            .iter()
+            .map(|index| {
+                BoundExpr::InputRef(BoundInputRef {
+                    index,
+                    return_type: input_types[index].clone(),
+                })
+            })
+            .collect();
+        LogicalProjection::new(exprs, self.clone_as_plan_ref()).into_plan_ref()
```
skyzh, line 133: Simply doing projection won't save I/Os. Will we have plan rules to merge `LogicalProjection` with `TableScan`, and merge `LogicalProjection` with `LogicalProjection`?

skyzh (approved):
Rest LGTM

src/optimizer/plan_nodes/mod.rs
st1page: Here we simply doing projection to maintain the correct behavior of the `fn prune_col`. in future, every PlanNode should implement its own `prune_col` method, which can call its inputs' `prune_col` recursively. in the end, when call the `prune_col` of the TableScan Node, it will remove the columns not required.

src/optimizer/mod.rs
st1page: Here we might need a collection which is more like a "Set", Maybe we will change it to [fixedbitset](https://crates.io/crates/fixedbitset)

PR #422, opened 2022-02-07
feat(cmd): interrupt long-running queries with ctrl-c.

arkbriar:
Supports the interruption of long-running queries with ctrl-c with `tokio::select` and `tokio::signal`.

**NOTE**: I haven't looked into the details but the future that `db.run` returns must be abort safe to avoid problems such as resource leaks. Let me know if it's not.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,6 +12,7 @@ use risinglight::storage::SecondaryStorageOptions;
 use risinglight::Database;
 use rustyline::error::ReadlineError;
 use rustyline::Editor;
+use tokio::{select, signal};
 use tracing::{info, warn, Level};
 use tracing_subscriber::prelude::*;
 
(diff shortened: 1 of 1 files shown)


skyzh (commented):
Thanks! This is really a handy feature. However, this is not abort-safe. If you run a query like:

```
select count(*) from lineitem;
```

You'll find that the storage thread panics at

```
                tx.send(result)
                    .await
                    .expect("failed to send chunk to compute thread");
```

and there is a warning, leading to leak of resource in transaction...

```
WARN risinglight::storage::secondary::transaction: Transaction dropped without committing or aborting
```

Not aborting a txn in RisingLight would cause issues in storage layer.

We might need to revisit all such places, so as to make this functionality really works.

arkbriar:
> Thanks! This is really a handy feature. However, this is not abort-safe. If you run a query like:
> 
> ```
> select count(*) from lineitem;
> ```
> 
> You'll find that the storage thread panics at
> 
> ```
>                 tx.send(result)
>                     .await
>                     .expect("failed to send chunk to compute thread");
> ```
> 
> and there is a warning, leading to leak of resource in transaction...
> 
> ```
> WARN risinglight::storage::secondary::transaction: Transaction dropped without committing or aborting
> ```
> 
> Not aborting a txn in RisingLight would cause issues in storage layer.
> 
> We might need to revisit all such places, so as to make this functionality really works.

Ahh, I see the panicked lines. It's a detached task without handling the close of channel. This one is simple but yes it's not as simple to cover all such places. I agree we should have a careful revisit.

Another problem you've mentioned is aborting the transaction. In current revision I find a non-async (but declared as async) method to abort the txn. Is it true that some more async things will be done whiling dealing with secondary storages in the future or we can just abort the running txn in `drop`?

skyzh:
Yes, we should delete flushed rowsets when aborting. Deleting files on disk is an async operation.

This operation can also be done in the background. For example, creating a vacuum thread, and simply send rowsets to cleanup through channel. Sending data to an unbounded channel doesn't require async. This would require a refactor.

skyzh:
I've created a tracking issue for graceful stop in https://github.com/risinglightdb/risinglight/issues/423. After the related parts have been refactored, we can then get this PR merged.

Issue #290, opened 2022-01-07
storage: merge next_batch_inner and next_batch_inner_with_filter in rowset_iterator


likg227:
next_batch_inner and next_batch_inner_with_filter are two separate functions now, but they have some duplicate parts and similar functionality. We hope these two functions can be merged into one function in the future.

PR #424, opened 2022-02-08
refactor(storage): merge next_batch_inner and next_batch_inner_with_filter in rowset_iterator

zzl200012:

Close: #290

### DIFF
--- a/src/storage/secondary/rowset/rowset_iterator.rs
+++ b/src/storage/secondary/rowset/rowset_iterator.rs
@@ -98,12 +98,7 @@ impl RowSetIterator {
         &mut self,
         expected_size: Option<usize>,
     ) -> StorageResult<(bool, Option<StorageChunk>)> {
-        let has_filter = self.filter_expr.is_some();
-        let filter_context = if self.filter_expr.is_some() {
-            Some(self.filter_expr.as_ref().unwrap())
-        } else {
-            None
-        };
+        let filter_context = self.filter_expr.as_ref();
         let fetch_size = if let Some(x) = expected_size {
             x
         } else {
@@ -130,10 +125,11 @@ impl RowSetIterator {
         // TODO: parallel fetch
         // TODO: align unmatched rows
 
+        let mut visibility_map = None;
+
         for (id, column_ref) in self.column_refs.iter().enumerate() {
-            if has_filter {
+            if let Some((_, filter_column)) = filter_context {
                 // Filter needed, so only filtered columns should be filled currently
-                let (_, filter_column) = filter_context.unwrap();
                 if !filter_column[id] {
                     arrays.push(None);
                     continue;
@@ -146,13 +142,24 @@ impl RowSetIterator {
                     if let Some((row_id, array)) = self.column_iterators[id]
                         .as_mut()
                         .unwrap()
-                        .next_batch(Some(fetch_size), None)
+                        .next_batch(Some(fetch_size), visibility_map.as_ref())
                         .await?
                     {
                         if let Some(x) = common_chunk_range {
                             if x != (row_id, array.len()) {
                                 panic!("unmatched rowid from column iterator");
                             }
+                        } else if !self.dvs.is_empty() {
+                            // Apply delete vector here to reduce the data
+                            // fetched next during scan
+                            let mut visi = BitVec::with_capacity(array.len());
+                            for _ in 0..array.len() {
+                                visi.push(true);
+                            }
+                            for dv in &self.dvs {
+                                dv.apply_to(&mut visi, row_id);
+                            }
+                            visibility_map = Some(visi);
                         }
                         common_chunk_range = Some((row_id, array.len()));
                         arrays.push(Some(array));
@@ -170,11 +177,10 @@ impl RowSetIterator {
             return Ok((true, None));
         };
 
-        let visibility = if has_filter {
+        let visible = if let Some((expr, _)) = filter_context {
             // Filter occurred, we should get the visibility map first by evaluating
             // expresstions on the filtered columns, and then retain the existing rows
             // for the other columns
-            let (expr, _) = filter_context.unwrap();
 
             // Need to optimize
             let bool_array = match expr
@@ -187,17 +193,16 @@ impl RowSetIterator {
 
             let mut filter_bitmap = BitVec::with_capacity(bool_array.len());
             for i in bool_array.iter() {
+                if let Some(visi) = visibility_map.as_ref() {
+                    if !visi[filter_bitmap.len()] {
+                        filter_bitmap.push(false);
+                        continue;
+                    }
+                }
                 if let Some(i) = i {
                     filter_bitmap.push(*i);
                 } else {
                     filter_bitmap.push(false);
-                }
-            }
-
-            // Apply delete vector to filter_bitmap
-            if !self.dvs.is_empty() {
-                for dv in &self.dvs {
-                    dv.apply_to(&mut filter_bitmap, common_chunk_range.0);
                 }
             }
 
@@ -220,22 +225,15 @@ impl RowSetIterator {
             }
 
             Some(filter_bitmap)
-        } else if self.dvs.is_empty() {
-            None
         } else {
-            let mut vis = BitVec::new();
-            vis.resize(common_chunk_range.1, true);
-            for dv in &self.dvs {
-                dv.apply_to(&mut vis, common_chunk_range.0);
-            }
-            Some(vis)
+            visibility_map
         };
 
-        if has_filter {
+        if filter_context.is_some() {
             // Use filter_bitmap to filter columns
             // TODO: Implement the skip interface for column_iterator and call it here.
             // For those already fetched columns, they also need to delete corrensponding blocks.
-            let filter_bitmap = visibility.as_ref().unwrap();
+            let filter_bitmap = visible.as_ref().unwrap();
             for (id, column_ref) in self.column_refs.iter().enumerate() {
                 match column_ref {
                     StorageColumnRef::RowHandler => continue,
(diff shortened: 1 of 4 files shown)


src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -94,12 +94,16 @@ impl RowSetIterator {
...
 
-    pub async fn next_batch_inner_with_filter(
+    pub async fn next_batch_inner(
         &mut self,
         expected_size: Option<usize>,
     ) -> StorageResult<(bool, Option<StorageChunk>)> {
-        let (expr, filter_column) = self.filter_expr.as_ref().unwrap();
-
+        let has_filter = self.filter_expr.is_some();
+        let filter_context = if self.filter_expr.is_some() {
```
skyzh, line 102: Looks the same as:

let filter_context = self.filter_expr.as_ref();

src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -126,26 +130,36 @@ impl RowSetIterator {
...
-                    .as_mut()
-                    .unwrap()
-                    .next_batch(Some(fetch_size), None)
-                    .await?
-                {
-                    if let Some(x) = common_chunk_range {
-                        if x != (row_id, array.len()) {
-                            panic!("unmatched rowid from column iterator");
+        for (id, column_ref) in self.column_refs.iter().enumerate() {
+            if has_filter {
```
skyzh, line 134: I would prefer using `if let Some(filter_context) = filter_context`, instead of combining `.is_some()` with `unwrap()`.

src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -156,165 +170,94 @@ impl RowSetIterator {
...
-                    StorageColumnRef::Idx(_) => {
-                        if arrays[id].is_none() {
-                            self.column_iterators[id]
-                                .as_mut()
-                                .unwrap()
-                                .skip(filter_bitmap.len());
-                        }
-                    }
+            // Apply delete vector to filter_bitmap
+            if !self.dvs.is_empty() {
```
skyzh, line 198: No need to check empty -- the for loop won't run if dvs is empty.

skyzh (commented):
Generally LGTM. In fact, we can do a further optimization:

* Applying delete vector and applying filtering expression are both a kind of filter scan.
* Therefore, we can firstly generate delete map.
* Then, apply filter map to the delete map.
* And finally, do filter scan.

Therefore, if there are a lot of rows being deleted, we can leverage filter scan to avoid scanning deleted rows.

zzl200012:
> Generally LGTM. In fact, we can do a further optimization:
> 
> * Applying delete vector and applying filtering expression are both a kind of filter scan.
> * Therefore, we can firstly generate delete map.
> * Then, apply filter map to the delete map.
> * And finally, do filter scan.
> 
> Therefore, if there are a lot of rows being deleted, we can leverage filter scan to avoid scanning deleted rows.

Updated. But I'm not sure if I've done this as you expected.

skyzh:
I'll take a look tomorrow, thanks for your contribution!

cc @likg227 might also take a look.

likg227:
LGTM！
It's better to write some unit tests for this pr.

src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -126,26 +125,48 @@ impl RowSetIterator {
...
+                            }
+                        } else if !self.dvs.is_empty() {
+                            // Apply delete vector here to reduce the data
+                            // fetched next during scan
+                            let mut visi = BitVec::with_capacity(array.len());
+                            for _ in 0..array.len() {
+                                visi.push(true);
+                            }
+                            for dv in &self.dvs {
+                                dv.apply_to(&mut visi, row_id);
```
skyzh, line 160: We can produce the delete vector visibility map before scanning the filter column? And we may only need to apply delete vector once :)

skyzh (commented):
I think the current approach doesn't seem very clear to me. What about applying the following idea in the scanning process?

* Firstly, we should determine a fetch_size (which has already been done in the current implementation.)
* After that, we will start to maintain a `visibility` `Option<BitVec>`. If this `visibility` variable is None, it means that at the current stage, we will scan all rows and won't skip any block.
* In the first step, we generate a visibility map from the deletion vector of the current rowset. If there is no deletion vector, `visibility` remains to be `None`.
* In the second step, we use the delete bitmap to scan the condition column in filter scan. (e.g., select * from table where a > 10, we scan the `a` column based on delete vector). If there are no filter conditions, we don't do any modification to the visibility map.
* Now we will get a visibility map: case 1. without filter condition, representing the delete vector; case 2. with filter condition, representing delete vector + filter bitmap.
* At this time, we can use this visibility map to scan: case 1. without filter condition, scan all columns; case 2. with filter condition, scanning all remaining columns.

I think in this way, we can make the code more understandable -- we always maintain a `visibility` bitmap, and modifying it little by little when scanning more and more columns.

If you have any question, feel free to discuss either at #risinglight-english or #risinglight-chinese channel.

Thanks for you contribution 🥰

src/storage/secondary/rowset/rowset_iterator.rs
zzl200012: I think we are on the same track, the only problem is that I didn't know how to get the start row_id in this batch without scanning 1 column first, which made some codes confusing... Here the `dv` would only be applied once as well, I guess. Maybe we could add an interface `get_current_row_id` for the column_iterator, and call it on the first column to generate the visibility map before doin scan? And after that, adjust the following code logic according to your comment, which would make the logic more clear.

skyzh: > Maybe we could add an interface get_current_row_id for the column_iterator

We definitely need a `current_row_id`, but there are multiple places we can add it.

> and call it on the first column

We can also record the `current_row_id` in `RowSetIterator`. See:

https://github.com/risinglightdb/risinglight/blob/53a3388ec5866290d6d2d01065121e41c372c025/src/storage/secondary/rowset/rowset_iterator.rs#L37-L40

Currently, the `current_row_id` always starts at 0.

You may choose the best place to add the `current_row_id` variable.

skyzh: After a second thought, maybe adding it on `ColumnIterator` is a better way.

zzl200012: Agree, since we already have https://github.com/risinglightdb/risinglight/blob/53a3388ec5/src/storage/secondary/column/concrete_column_iterator.rs#L46, the only work is to encapsulate a method in `ColumnIterator `

skyzh: It seems that we already have `get_current_row_id`, but only exposed for `test` 🤪

zzl200012:
Reworked the code just now, PTAL when you have time~ cc @skyzh

src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -126,189 +125,157 @@ impl RowSetIterator {
...
+        // we would scan all rows and won't skip any block.
+        let mut visibility_map = None;
+
+        // Generate the initial visibility_map from delete vectors, so
+        // that we can avoid unnecessary scan on filter column at next
+        if !self.dvs.is_empty() {
+            // Get the start row id first
+            let start_row_id = {
+                let mut row_id = 0;
+                for (id, column_ref) in self.column_refs.iter().enumerate() {
```
skyzh, line 141: All column iterators will have the same row_id at any time -- I think it's okay to simple call `fetch_current_row_id` for the `0` position column iterator.

src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -126,189 +125,157 @@ impl RowSetIterator {
...
                 }
-            } else {
-                arrays.push(None);
+                row_id
+            };
+
+            // Initialize visibility map and apply delete vector to it
+            let mut visi = BitVec::new();
+            visi.resize(fetch_size, true);
+            for dv in &self.dvs {
```
skyzh, line 159: The `visi` can remain `None` if `self.dvs` is empty. We can do this optimization later.

src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -126,189 +125,157 @@ impl RowSetIterator {
...
+            let bool_array = match expr
+                .eval_array_in_storage(&arrays, common_chunk_range.1)
+                .unwrap()
+            {
+                ArrayImpl::Bool(a) => a,
+                _ => panic!("filters can only accept bool array"),
+            };
+
+            let mut filter_bitmap = BitVec::with_capacity(bool_array.len());
+            for i in bool_array.iter() {
```
skyzh, line 212: .enumerate()

src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -126,189 +125,157 @@ impl RowSetIterator {
...
+                if let Some(visi) = visibility_map.as_ref() {
+                    if !visi[filter_bitmap.len()] {
```
skyzh, line 214: (bug) should use the index of `bool_array` instead of `.len()`?

skyzh (commented):
Despite the bug, rest LGTM! Other minor fixes can be done in a separate PR (if you want). Thanks for this great work.

skyzh:
... also we can remove the original `get_current_row_id` function (maybe in another PR?)

zzl200012:
> ... also we can remove the original `get_current_row_id` function (maybe in another PR?)

Alright, I would start a new PR tomorrow to do the stuff above

skyzh (approved):

PR #427, opened 2022-02-09
feat(planner): resolve schema in plan node

### LINKED ISSUE #245: proposal(catalog): move name field to `ColumnDesc` definition instead of `ColumnCatalog`
```rust
pub struct ColumnDesc {
    datatype: DataType,
    is_primary: bool,
}
```
and 
```rust
pub struct ColumnCatalog {
    id: ColumnId,
    name: String,
    desc: ColumnDesc,
}
```
should it be like this? we can move the column name to the `ColumnDesc`.
that makes sense, and we can use ColumnDesc in the schema of `PlanNode` and `Executor`, so that we can output the column name in the query result. #231 
```rust
pub struct ColumnDesc {
    datatype: DataType,
    name: String,
    is_primary: bool,
}
pub struct ColumnCatalog {
    id: ColumnId,
    desc: ColumnDesc,
}
```

pleiadesian:
Close #245 

Signed-off-by: Zulai Wang <zl31wang@gmail.com>

### DIFF
--- a/src/optimizer/plan_nodes/logical_copy_from_file.rs
+++ b/src/optimizer/plan_nodes/logical_copy_from_file.rs
@@ -18,13 +18,21 @@ pub struct LogicalCopyFromFile {
     format: FileFormat,
     /// The column types.
     column_types: Vec<DataType>,
+    /// Schema of the plan node
+    schema: Vec<ColumnDesc>,
 }
 impl LogicalCopyFromFile {
-    pub fn new(path: PathBuf, format: FileFormat, column_types: Vec<DataType>) -> Self {
+    pub fn new(
+        path: PathBuf,
+        format: FileFormat,
+        column_types: Vec<DataType>,
+        schema: Vec<ColumnDesc>,
+    ) -> Self {
         Self {
             path,
             format,
             column_types,
+            schema,
         }
     }
 
@@ -46,8 +54,8 @@ impl LogicalCopyFromFile {
 impl PlanTreeNodeLeaf for LogicalCopyFromFile {}
 impl_plan_tree_node_for_leaf!(LogicalCopyFromFile);
 impl PlanNode for LogicalCopyFromFile {
-    fn out_types(&self) -> Vec<DataType> {
-        self.column_types.clone()
+    fn schema(&self) -> Vec<ColumnDesc> {
+        self.schema.clone()
     }
 }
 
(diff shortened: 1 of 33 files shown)


src/optimizer/plan_nodes/logical_copy_from_file.rs
```diff
@@ -18,13 +18,21 @@ pub struct LogicalCopyFromFile {
     format: FileFormat,
     /// The column types.
     column_types: Vec<DataType>,
```
wangrunji0408, line 20: Now this field is redundant.
```suggestion
```

src/optimizer/plan_nodes/logical_values.rs
```diff
@@ -13,13 +13,19 @@ use crate::types::DataType;
 #[derive(Debug, Clone, Serialize)]
 pub struct LogicalValues {
     column_types: Vec<DataType>,
```
wangrunji0408, line 15: Now this field is redundant.
```suggestion
```

src/optimizer/plan_nodes/physical_simple_agg.rs
```diff
@@ -44,8 +35,22 @@ impl PlanTreeNodeUnary for PhysicalSimpleAgg {
...
+            .map(|agg_call| {
+                use crate::binder::AggKind::*;
+                let name = match agg_call.kind {
+                    Avg => "avg",
+                    RowCount | Count => "count",
+                    Max => "max",
+                    Min => "min",
+                    Sum => "sum",
+                }
+                .to_string();
```
wangrunji0408, line 50: What about:
```suggestion
                let name = format!("{:?}", agg_call.kind);
```

src/optimizer/plan_nodes/logical_projection.rs
```diff
@@ -50,10 +50,25 @@ impl PlanTreeNodeUnary for LogicalProjection {
...
-    fn out_types(&self) -> Vec<DataType> {
+    fn schema(&self) -> Vec<ColumnDesc> {
+        let child_schema = self.child.schema();
         self.project_expressions
             .iter()
-            .map(|expr| expr.return_type().unwrap())
+            .map(|expr| {
+                let name = match expr {
+                    BoundExpr::ColumnRef(column_ref) => column_ref.desc.name().to_string(),
+                    BoundExpr::TypeCast(type_cast) => match expr {
```
wangrunji0408, line 60:
```suggestion
                    BoundExpr::TypeCast(type_cast) => match type_cast.expr {
```

wangrunji0408 (commented):
LGTM!

src/optimizer/plan_nodes/logical_copy_from_file.rs
pleiadesian: It is used by `column_types()`, which returns the reference to `column_types` to caller. So, I think redundantly maintaining a `column_types` field is acceptable.

wangrunji0408 (approved):

PR #429, opened 2022-02-09
chore: add column name to output

### LINKED ISSUE #231: binder: show column name in query result
```
> \dt
+---+----------+---+----------+---+---+
| 0 | postgres | 0 | postgres | 0 | t |
+---+----------+---+----------+---+---+

> select * from t;
+---+----+-----+
| 3 | 30 | 300 |
| 4 | 40 | 400 |
| 3 | 30 | 300 |
| 4 | 50 | 500 |
+---+----+-----+
```

We should display column name in query result, e.g., `v1, v2, v3`.

pleiadesian:
Close #231 

This PR depends on #427

### DIFF
--- a/src/optimizer/plan_nodes/logical_aggregate.rs
+++ b/src/optimizer/plan_nodes/logical_aggregate.rs
@@ -89,3 +89,46 @@ impl fmt::Display for LogicalAggregate {
         writeln!(f, "LogicalAggregate: {} agg calls", self.agg_calls.len(),)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::binder::AggKind;
+    use crate::types::{DataTypeExt, DataTypeKind};
+
+    #[test]
+    fn test_aggregate_out_names() {
+        let plan = LogicalAggregate::new(
+            vec![
+                BoundAggCall {
+                    kind: AggKind::Sum,
+                    args: vec![],
+                    return_type: DataTypeKind::Double.not_null(),
+                },
+                BoundAggCall {
+                    kind: AggKind::Avg,
+                    args: vec![],
+                    return_type: DataTypeKind::Double.not_null(),
+                },
+                BoundAggCall {
+                    kind: AggKind::Count,
+                    args: vec![],
+                    return_type: DataTypeKind::Double.not_null(),
+                },
+                BoundAggCall {
+                    kind: AggKind::RowCount,
+                    args: vec![],
+                    return_type: DataTypeKind::Double.not_null(),
+                },
+            ],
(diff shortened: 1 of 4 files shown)


wangrunji0408 (commented):
LGTM!
We will merge this after #427.

skyzh (approved):
I think it's better not to bind `header` with data chunk. 

* Not all data chunks have headers. For example, we might store DataChunks in in-memory engine.
* Every data chunk has its own header, while the output of one executor should have the same header. The header is duplicated multiple times, and it is possible that we forget to attach header in some chunks.

I'm okay with the current approach, but I think it's better to store `header` in another way.

skyzh:
... by the way, would you please add some unit tests for plan node out name?

src/optimizer/plan_nodes/logical_aggregate.rs
```diff
@@ -89,3 +89,46 @@ impl fmt::Display for LogicalAggregate {
...
+                    args: vec![],
+                    return_type: DataTypeKind::Double.not_null(),
+                },
+            ],
+            vec![],
+            Arc::new(Dummy {}),
+        );
+
+        let column_names = plan.out_names();
+        assert_eq!(column_names[0], "sum");
```
skyzh, line 129: The column name should be something like `sum_x`, instead of simply sum?

skyzh (approved):
LGTM, good work! We can later improve the deduced column name to be more meaningful (Please create a tracking issue for that. It may take some effort to print something like `a.x` in join plans and `sum_x` in agg.) We can get this merged first!

src/optimizer/plan_nodes/logical_aggregate.rs
pleiadesian: It's consistent with pg's output, which is simply sum.
```
mydb=# select sum(v1), count(v2) from t;
 sum | count
-----+-------
   1 |     1
```

pleiadesian:
> LGTM, good work! We can later improve the deduced column name to be more meaningful (Please create a tracking issue for that. It may take some effort to print something like `a.x` in join plans and `sum_x` in agg.) We can get this merged first!

All output of column names in current PR is refered from postgresql's output. So, I think we don't have to do more deduction on column names.

skyzh:
> > LGTM, good work! We can later improve the deduced column name to be more meaningful (Please create a tracking issue for that. It may take some effort to print something like `a.x` in join plans and `sum_x` in agg.) We can get this merged first!
> 
> All output of column names in current PR is refered from postgresql's output. So, I think we don't have to do more deduction on column names.

Thanks for investigating into this!

PR #438, opened 2022-02-10
chore(storage): remove redundant get_current_row_id

zzl200012:

Follow-up of #424 
For [the optimization](https://github.com/risinglightdb/risinglight/pull/424#discussion_r802473971), I think we already ensured `self.dvs` not to be empty at L137, right? cc @skyzh
https://github.com/risinglightdb/risinglight/blob/1a026b6ff74c57b27688c904094f145861931bf4/src/storage/secondary/rowset/rowset_iterator.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L135-L153


skyzh (approved):
Thanks! You're right.

Issue #434, opened 2022-02-09
cli: support human-readable message for create, insert and explain


skyzh:
Currently, when we are creating new tables and inserting things, we will see:

```
> create table t2(v1 int not null, v2 int not null);
+---+
| 0 |
+---+
```

The `0` is the return result (as data chunk) from the create table executor.

We should be able to make this into human-readable format. We can do this with our new column name framework.

* InsertExecutor should return a column with special name `$insert.row_counts`.
* When cli finds the special column name, the data chunk will be turned into a human-readable message: `x rows inserted`.

The same applies to the explain statement. We should directly print the explain result string, instead of wrapping a data chunk frame around the output.

PR #440, opened 2022-02-10
human-readable message for create & insert & explain

xxchan:
closes https://github.com/risinglightdb/risinglight/issues/434

image

### DIFF
--- a/src/array/data_chunk.rs
+++ b/src/array/data_chunk.rs
@@ -119,7 +119,15 @@ impl fmt::Display for DataChunk {
         use comfy_table::Table;
         let mut table = Table::new();
         if let Some(header) = &self.header {
-            table.set_header(header);
+            match header[0].as_str() {
+                "$insert.row_counts" => {
+                    return write!(f, "{} rows inserted", self.arrays[0].get_to_string(0))
+                }
+                "$create" => return write!(f, "created"),
+                _ => {
+                    table.set_header(header);
+                }
+            }
         }
         table.load_preset("||--+-++|    ++++++");
         for i in 0..self.cardinality() {
--- a/src/array/data_chunk.rs
+++ b/src/array/data_chunk.rs
@@ -111,6 +111,10 @@ impl DataChunk {
     pub fn set_header(&mut self, header: Vec<String>) {
         self.header = Some(header);
     }
+
+    pub fn header(&self) -> Option<&[String]> {
+        self.header.as_deref()
+    }
 }
 
 /// Print the chunk as a pretty table.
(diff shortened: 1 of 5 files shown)


xxchan:
It seems that explain needs to change `fmt` impl of `PhysicalInsert` & `PhysicalCreateTable`. I'm not sure whether it's proper 🤔.

src/array/data_chunk.rs
```diff
@@ -119,7 +119,15 @@ impl fmt::Display for DataChunk {
...
-            table.set_header(header);
+            match header[0].as_str() {
+                "$insert.row_counts" => {
+                    return write!(f, "{} rows inserted", self.arrays[0].get_to_string(0))
+                }
+                "$create" => return write!(f, "created"),
+                _ => {
+                    table.set_header(header);
+                }
+            }
```
wangrunji0408, line 130: Yes I also find it a little weird to put this code in `Display`. What about moving it to main.rs?

https://github.com/risinglightdb/risinglight/blob/5b4af83bcd50592d5675351bd093c12908c31f58/src/main.rs#L58

wangrunji0408:
> It seems that explain needs to change `fmt` impl of `PhysicalInsert` & `PhysicalCreateTable`. I'm not sure whether it's proper 🤔.

If I understand correctly, what we want to do for EXPLAIN is:
```diff
  > explain create table tt (x int, y int);
- +---------------------------------------------------------------------------------+
- | PhysicalCreateTable: table tt, columns [x:Int(None) (null), y:Int(None) (null)] |
- |                                                                                 |
- +---------------------------------------------------------------------------------+
+ PhysicalCreateTable: table tt, columns [x:Int(None) (null), y:Int(None) (null)]
```

So it only needs a special column name, for example `$explain`, and a special print like you did for INSERT. No need to touch the plans? 🤔

src/array/data_chunk.rs
xxchan: I put it here since there are multiple entry points (`interactive`, `run_sql`, `run_sqllogictest`). And putting here seems ok semantically (`Display` for displaying pretty message). I think if we leave `Display` unchanged, will the dummy pretty tables for `create`, `insert` be otherwise meaningful? 

But I don't have a strong opinion. Both are fine for me.

xxchan:
Yes, I just found that I misunderstood the meaning of changing explain's message. I thought it means to change the explain of "insert" and "create" 😂😂😂  

So I will finish explain similarly

src/array/data_chunk.rs
wangrunji0408: Got your idea. But I think these messages are CLI-specific, and `Display for DataChunk` should show the original structure of a chunk. Currently we can not distinguish between the following 2 chunks 🙈:
```
+---------+  +---------+
| $create |  | created |
+---------+  +---------+
| 0       |
+---------+
```
So I still prefer to move these code into CLI. We can put them in a function if used in multiple places.

wangrunji0408:
Good work and good night! 🌚

wangrunji0408 (approved):
LGTM!
Thanks for your contribution!

Issue #362, opened 2022-01-26
executor: hash join support multi columns as hash key.


st1page:
Now, our hash join just support use one column pair as the hash key, it can support multi columns as hash key.

st1page:
-  when the [`LogicalJoin`](https://github.com/risinglightdb/risinglight/blob/main/src/optimizer/plan_nodes/logical_join.rs) constructed, [`JoinPredicate`](https://github.com/risinglightdb/risinglight/blob/main/src/optimizer/plan_nodes/join_predicate.rs#L12) has analyzed the join on clause condition and get the equal keys of the inputRef and as a field at `LogicalJoin`.
- [`PhysicalHashJoin`](https://github.com/risinglightdb/risinglight/blob/main/src/optimizer/plan_nodes/physical_hash_join.rs#L11) does not use the `self.logical.predicate()`, but use its own index fields now. you may need to remove them and use eq_keys in JoinPredicate.
  - `LogicalJoin` is converted to `PhysicalHashJoin` [here](https://github.com/risinglightdb/risinglight/blob/main/src/optimizer/logical_plan_rewriter/convert_physical.rs#L43)
- there is only one index in both two side in [`HashJoinExecutor`](https://github.com/risinglightdb/risinglight/blob/main/src/executor/hash_join.rs#L20) now. you need to change the fields and rewrite the logic in `fn execute(self)`
  - HashJoinExecutor is built from `PhysicalHashJoin` planNode [here](https://github.com/risinglightdb/risinglight/blob/main/src/executor/mod.rs#L274)

PR #443, opened 2022-02-11
executor: hash join support multi columns as hash key.

xxchan:
closes https://github.com/risinglightdb/risinglight/issues/362

- `PhysicalHashJoin`'s `left/right_column_index` fields are deleted, only `logical` left. 
- `HashJoinExecutor`'s same fields are changed from `index` -> `indexes`. They are computed in `visit_physical_hash_join` now. (Previously in `rewrite_logical_join`)

### DIFF
--- a/src/array/data_chunk.rs
+++ b/src/array/data_chunk.rs
@@ -182,6 +182,13 @@ impl RowRef<'_> {
         self.chunk.array_at(idx).get(self.row_idx)
     }
 
+    pub fn get_by_indexes(&self, indexes: &[usize]) -> Vec<DataValue> {
+        indexes
+            .iter()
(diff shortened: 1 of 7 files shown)


skyzh (commented):
Generally LGTM, better to add some e2e test cases in https://github.com/risinglightdb/risinglight/blob/main/tests/sql/join.slt?

btw, could you provide some examples of plans after this change?

xxchan:
main branch:

image


---

Last version: There's unnecessary Eq in the pulled filter! I should have tested it carefully 😭

image

--- 

Current version:

image

xxchan:
Swapping LHS and RHS of equations works correctly

image

skyzh (approved):

st1page (approved):
optimizer part LGTM

PR #445, opened 2022-02-11
feat(optimizer): add cardinality trait

MingjiHan99:
Signed-off-by: MingjiHan <mingji_han@outlook.com>

We may need an execution context (like DuckDB) to provide statitsics in storage layer for optimizer @skyzh .

### DIFF
--- a/src/optimizer/plan_nodes/logical_limit.rs
+++ b/src/optimizer/plan_nodes/logical_limit.rs
@@ -32,6 +32,10 @@ impl LogicalLimit {
     pub fn limit(&self) -> usize {
         self.limit
     }
+    #[allow(dead_code)]
+    fn estimated_cardinality(&self) -> usize {
+        self.limit
+    }
 }
 impl PlanTreeNodeUnary for LogicalLimit {
     fn child(&self) -> PlanRef {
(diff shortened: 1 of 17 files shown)


src/optimizer/plan_nodes/logical_limit.rs
```diff
@@ -32,6 +32,10 @@ impl LogicalLimit {
     pub fn limit(&self) -> usize {
         self.limit
     }
+    #[allow(dead_code)]
+    fn estimated_cardinality(&self) -> usize {
```
skyzh, line 36: should be implemented on `PlanNode`?

src/optimizer/plan_nodes/physical_projection.rs
```diff
@@ -21,6 +21,10 @@ impl PhysicalProjection {
     pub fn logical(&self) -> &LogicalProjection {
         &self.logical
     }
+    #[allow(dead_code)]
+    fn estimated_cardinality(&self) -> usize {
```
skyzh, line 25: ditto?

skyzh (approved):
Rest LGTM, good work! It's true that we might expose the storage context to the optimizer, so that we can call storage's statistics interface.

src/optimizer/plan_nodes/logical_limit.rs
xxchan: Why this comment is not resolved...?

Issue #444, opened 2022-02-11
executor: limit returns nothing if offset is out of bound


xxchan:
Currently it returns no chunks. If instead an empty chunk is returned, the table header can be shown, which is more user-friendly. (I think this can be a good-first-issue 😄

![image](https://user-images.githubusercontent.com/37948597/153642640-b0718cf2-96fd-4336-8a66-2929957a7e63.png)

skyzh:
I think a better solution is to make header part of the query result (instead of part of the data chunk), so that the caller can:

* `.run(sql: &str) -> (ColumnDesc, Vec<DataChunk>)`
* where the first parameter is the description of columns
* and the data chunk don't have to add a `header` field.

xxchan:
I agree! Then do we rename this issue or.create a new issue？

skyzh:
> I agree! Then do we rename this issue or.create a new issue？

We already have one 🥳😇 https://github.com/risinglightdb/risinglight/issues/435

Issue #404, opened 2022-02-03
executor: support `TopN`


skyzh:
When running TPC-H queries, it is possible that we have `limit` with `order by`. We can optimize this into `TopN`.

- [ ] add `TopN` executor. Simply create a new executor in https://github.com/risinglightdb/risinglight/tree/main/src/executor and add unit tests. The TopN executor should consume chunks from children, find topN of a column, and return the full row to the parent executor. Note that by saying top N, it means that top offset .. limit items.
- [ ] add optimization rules to rewrite `limit + order by` to `TopN`. We will need to add new plan nodes in https://github.com/risinglightdb/risinglight/tree/main/src/optimizer.

TennyZhuang:
The [binary-heap-plus](https://crates.io/crates/binary-heap-plus) crate may be helpful to the issue with the comparator-api support.

skyzh:
> The [binary-heap-plus](https://crates.io/crates/binary-heap-plus) crate may be helpful to the issue with the comparator-api support.

That's cool! I originally thought that we might need to hand-write a binary heap as in MergeIterator...

https://github.com/risinglightdb/risinglight/blob/a2fd396051263f5ace8c0dbb85c576f680008c00/src/storage/secondary/merge_iterator.rs#L11-L34

xxchan:
I'd like to try this

PR #446, opened 2022-02-11
executor: support TopN

xxchan:
closes https://github.com/risinglightdb/risinglight/issues/404

also closes https://github.com/risinglightdb/risinglight/issues/444 (let limit return a dummy chunk if offset is out of bound or limit=0)
(since I tested `TopN` using `Limit`

Some example runs:

image

### DIFF
--- a/src/array/data_chunk.rs
+++ b/src/array/data_chunk.rs
@@ -115,6 +115,18 @@ impl DataChunk {
     pub fn header(&self) -> Option<&[String]> {
         self.header.as_deref()
     }
+
+    pub fn from_rows<'a>(rows: &[RowRef<'a>], chunk: &Self) -> Self {
+        let mut arrays = vec![];
+        for col_idx in 0..chunk.column_count() {
+            let mut builder = ArrayBuilderImpl::from_type_of_array(chunk.array_at(col_idx));
+            for row in rows {
+                builder.push(&row.get(col_idx));
+            }
+            arrays.push(builder.finish());
+        }
+        arrays.into_iter().collect()
+    }
 }
 
 /// Print the chunk as a pretty table.
--- a/src/executor/top_n.rs
+++ b/src/executor/top_n.rs
@@ -1,0 +1,161 @@
+// Copyright 2022 RisingLight Project Authors. Licensed under Apache-2.0.
+
+use std::cmp::Ordering;
+
+use binary_heap_plus::BinaryHeap;
+
+use super::*;
+use crate::array::DataChunk;
+use crate::binder::BoundOrderBy;
+
+/// The executor of a Top N operation.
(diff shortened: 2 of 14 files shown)


src/executor/top_n.rs
```diff
@@ -0,0 +1,161 @@
...
+        let mut heap = BinaryHeap::with_capacity_by(heap_size, |row1, row2| {
+            cmp(row1, row2, &self.comparators)
+        });
+
+        // collect all chunks
+        let mut chunks = vec![];
+        #[for_await]
+        for batch in self.child {
+            chunks.push(batch?);
+        }
```
xxchan, line 32: I wanted to update heap inside the loop here, but it seems impossible since the rows are references (so we cannot drop the batches).🤔

src/executor/top_n.rs
```diff
@@ -0,0 +1,160 @@
...
+        [ArrayImpl::Int32(array)].into_iter().collect()
+    }
+
+    fn equivalent_executors(
+        inputs: Vec<DataChunk>,
+        offset: usize,
+        limit: usize,
+        catalog: Vec<ColumnCatalog>,
+        idx_desc: Vec<(usize, bool)>,
+    ) -> (TopNExecutor, LimitExecutor) {
```
xxchan, line 126: Actually I'm considering randomized/property-based testing.🤔 (Maybe we can do it in the future)

src/array/data_chunk.rs
```diff
@@ -115,6 +115,18 @@ impl DataChunk {
     pub fn header(&self) -> Option<&[String]> {
         self.header.as_deref()
     }
+
+    pub fn from_rows<'a>(rows: &[RowRef<'a>], chunk: &Self) -> Self {
```
skyzh, line 119: I prefer to pass a schema here, instead of a chunk. Add a TODO here?

src/executor/top_n.rs
skyzh: I think it's okay to store rows with ownership inside the heap?

xxchan: Is there a type for owned rows? I only find `RowRef`

st1page (approved):
LGTM

src/executor/top_n.rs
xxchan: Oh, do you mean `rows.values().collect`: `Vec<DataValue>`

src/array/data_chunk.rs
xxchan: Yes I have the same idea. It will be more elegant 🥰
```suggestion
    /// TODO: pass a schema here, instead of a chunk
    pub fn from_rows<'a>(rows: &[RowRef<'a>], chunk: &Self) -> Self {
```

Issue #428, opened 2022-02-09
storage: support blob encoding


skyzh:
After https://github.com/risinglightdb/risinglight/pull/386 gets merged, we should also be able to store blob on disk.

My proposal is to migrate `varchar_block` to store both blob type and varchar type.

PR #461, opened 2022-02-14
storage: support blob encoding on disk

xxchan:
closes https://github.com/risinglightdb/risinglight/issues/428

Actually I'm not very sure about how to organize the code (so feel free to completely overthrow my idea!).
The question is: for each of block/column builder/iterator, whether to put varchar & blob together or put varchar & char together? 

Here's the summary of current changes of mine:
- block builder & iterator:

  `PlainVarcharBlockBuilder/Iterator` are changed into 
  `PlainBlobBlockBuilder/Iterator<T: BlobEncode + ?Sized>`. 

  `BlobEncode` is an imitation of `PrimitiveFixedWidthEncode`
- column builder & iterator: directly copied from char's (with fixed char's part removed)

Currently the test failed because
```
thread 'tokio-runtime-worker' panicked at 'no user column', src/storage/secondary/rowset/rowset_iterator.rs:56:13
```

This is caused by `DELETE FROM blobs WHERE true`, which should be fixed in https://github.com/risinglightdb/risinglight/issues/421


src/storage/secondary/encode.rs
```diff
@@ -112,3 +115,47 @@ impl PrimitiveFixedWidthEncode for Interval {
...
+
+impl Len for str {
+    fn len(&self) -> usize {
+        self.len()
+    }
+}
+
+pub trait FromBytes {
+    fn from_bytes(bytes: &[u8]) -> &Self;
+}
```
xxchan, line 149: I tried `From<&[u8]>` but it doesn't work because of lifetime... Is there a better way?

```rust
impl From<&[u8]> for &BlobRef {
    fn from(bytes: &[u8]) -> Self {
        BlobRef::new(bytes)
    }
}

error[E0495]: cannot infer an appropriate lifetime for lifetime parameter in function call due to conflicting requirements
  --> src/types/blob.rs:20:9
   |
20 |         BlobRef::new(bytes)
   |         ^^^^^^^^^^^^^^^^^^^
   |
note: first, the lifetime cannot outlive the anonymous lifetime defined here...
  --> src/types/blob.rs:19:20
   |
19 |     fn from(bytes: &[u8]) -> Self {
   |                    ^^^^^
note: ...so that reference does not outlive borrowed content
  --> src/types/blob.rs:20:22
   |
20 |         BlobRef::new(bytes)
   |                      ^^^^^
note: but, the lifetime must be valid for the lifetime `'_` as defined here...
  --> src/types/blob.rs:18:11
   |
18 | impl From<&[u8]> for &BlobRef {
   |           ^
note: ...so that reference does not outlive borrowed content
  --> src/types/blob.rs:20:9
   |
20 |         BlobRef::new(bytes)
   |         ^^^^^^^^^^^^^^^^^^^

```

skyzh: Should be `impl <'a> From<&[u8]> for BlobRef<'a>`

skyzh: That's the most confusing part of Rust lifetime 🤪 I'll review this PR later.

xxchan: If so, it will have problem in implementing `BlockIterator`

```rust
impl<'a, T: BlobEncode + ?Sized + From<&'a [u8]>> BlockIterator<T::ArrayType>
    for PlainBlobBlockIterator<T>
{
    fn next_batch(
        &mut self,
        expected_size: Option<usize>,
        builder: &mut <T::ArrayType as Array>::Builder,
    ) -> usize {
		// ...
		
        // cannot infer an appropriate lifetime for lifetime parameter in function call due to conflicting requirements
		// one is &mut self, one is 'a
        let data_buffer = &self.block[offsets_length..]; 
		
		// ...
		builder.push(Some(&T::from(&data_buffer[from..to])));
}
```

Is there a way to handle this problem?

skyzh: We will need HRTB here 🤪

```rust
impl<'a, T> BlockIterator<T::ArrayType>
    for PlainBlobBlockIterator<T> where T: BlobEncode + ?Sized, for<'a> T::Item<'a>: From<&'a [u8]>
```

xxchan: 🤯 got it

xxchan: Oh wait, why don't I write these bounds as methods in `BlobEncode`....

src/types/blob.rs
```diff
@@ -115,6 +115,12 @@ impl AsRef<[u8]> for BlobRef {
     }
 }
 
+impl AsRef<BlobRef> for [u8] {
```
skyzh, line 118: I think it's better not to implement `AsRef` for very commonly-used types like `[u8]`. `to_byte_slice` and `from_byte_slice` all should be part of `BlobEncode`.

src/storage/secondary/column/blob_column_builder.rs
```diff
@@ -0,0 +1,75 @@
...
+use risinglight_proto::rowset::block_index::BlockType;
+use risinglight_proto::rowset::BlockIndex;
+
+use super::super::{BlockBuilder, BlockIndexBuilder, PlainBlobBlockBuilder};
+use super::{append_one_by_one, ColumnBuilder};
+use crate::array::{Array, BlobArray};
+use crate::storage::secondary::ColumnBuilderOptions;
+use crate::types::BlobRef;
+
+/// Column builder of char types.
```
skyzh, line 12:
```suggestion
/// Column builder of blob types.
```

skyzh (commented):
LGTM, good work!

src/storage/secondary/encode.rs
skyzh: My fault, it seems that `BlobRef` doesn't have a lifetime parameter. So we won't need to experiment with the above code :(

skyzh:
> Actually I'm not very sure about how to organize the code (so feel free to completely overthrow my idea!).
The question is: for each of block/column builder/iterator, whether to put varchar & blob together or put varchar & char together?

The current organization looks good enough.

> Currently the test failed because
> ```
> ---- disk::_blob_slt_ stdout ----
> thread 'tokio-runtime-worker' panicked at 'no user column', src/storage/secondary/rowset/rowset_iterator.rs:56:13
> ```

I think we can temporarily disable the test cases and create an issue for that, so as to get this PR merged at the first place.

Thanks!

src/types/blob.rs
xxchan: Yes. Btw, this impl seems useless. I forget why I added it 😅

skyzh (approved):

Issue #348, opened 2022-01-21
executor: refactor executors to split chunks using DataChunkBuilder


TennyZhuang:
- [x] #349 
- [x] #466
- [x] #523 
- [x] #607
- [x] #618
- [x] #619

skyzh:
Refactoring executors to split chunk seems like a good-first-issue. If you are interested in refactoring one of our executors, feel free to create a tracking issue!

wangqiim:
I think it don't need introduce DataChunkBuilder to TopN. Is it right?

skyzh:
I think we need to. TopN might produce very large chunk (e.g., offset 0 limit 10000). It's better to get it chunked, so as to make the parent executor easier to handle.

skyzh:
I've created an issue for that.

xxchan:
All done!

PR #462, opened 2022-02-14
feat(executor): split chunks in HashJoinExecutor

TennyZhuang:

Part of #348 

It seems that using macro is the only solution.


skyzh (commented):
I would prefer implementing a data chunk builder :)

wangrunji0408 (approved):
LGTM.
Also +1 for data chunk builder.

skyzh:
I prefer to keep this PR on hold instead of merging. Prefer the chunk builder implementation.

PR #465, opened 2022-02-14
refactor(storage): remove filter scan functionality on column iterator

### LINKED ISSUE #452: storage: remove filter scan functionality on column iterator
In column iterator, we have implemented...

https://github.com/risinglightdb/risinglight/blob/c221b085ceb4a61fe1cf1e6c9d6caa36b7215a2e/src/storage/secondary/column/concrete_column_iterator.rs#L137-L141

This function will only have effect (reduce I/O) when a single call of `next_batch` spans across multiple blocks. However, this is a very rare case.

ColumnIterator is used by RowSetIterator. In RowSetIterator, we already have a very fine-grained control of how many rows to scan:

https://github.com/risinglightdb/risinglight/blob/c221b085ceb4a61fe1cf1e6c9d6caa36b7215a2e/src/storage/secondary/rowset/rowset_iterator.rs#L102-L120

Column iterators' fetch hint is simply remaining items of the current block. Therefore, we seldom perform a single `next_batch` call across multiple blocks.

Therefore, we can solely leverage `next_batch` (w/o filter map) and `skip` to perform filter scan on columns. We call `.skip` only when the computed filter bitmap is all 0.

- [ ] refactor RowSetIterator to never use visibility map on column
- [ ] set fetch size to `min(expected_size, fetch_size)`. (Currently the table scan executor only calls `next_batch` with None. But we should prepare for future cases.)
  * We can also call `next_batch` multiple times until we reach `expected_size`. Not important.
- [ ] remove `next_batch_with_filter` on `ColumnIterator`.

cc @likg227 any ideas?

zzl200012:

close #452

### DIFF
--- a/src/storage/secondary/column/concrete_column_iterator.rs
+++ b/src/storage/secondary/column/concrete_column_iterator.rs
@@ -196,7 +196,9 @@ impl<A: Array, F: BlockIteratorFactory<A>> ConcreteColumnIterator<A, F> {
         assert_eq!(cnt, 0);
 
         // indicate that a new block (located by `current_block_id`) should be
-        // loaded at the beginning of the next `next_batch`
+        // loaded at the beginning of the next `next_batch`. We reserve this field
+        // since `skip` may be called several times, and we need it to correctly
+        // compute remaining items during skipping.
         self.is_fake_iter = true;
     }
 }
--- a/src/storage/secondary/column/concrete_column_iterator.rs
+++ b/src/storage/secondary/column/concrete_column_iterator.rs
@@ -1,8 +1,5 @@
 // Copyright 2022 RisingLight Project Authors. Licensed under Apache-2.0.
 
-use std::cmp::min;
-
-use bitvec::prelude::BitVec;
 use futures::Future;
 use risinglight_proto::rowset::block_index::BlockType;
 use risinglight_proto::rowset::BlockIndex;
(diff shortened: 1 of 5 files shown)


likg227:
LGTM!
It's seems that you have done a great job, but there are some small problems. Also, how about remove FakeBlockIterator?

src/storage/secondary/column/concrete_column_iterator.rs
```diff
@@ -275,29 +191,17 @@ impl<A: Array, F: BlockIteratorFactory<A>> ConcreteColumnIterator<A, F> {
...
         assert_eq!(cnt, 0);
 
-        self.is_fake_iter = true;
-        self.block_iterator = self.factory.get_fake_iterator(
-            self.column.index().index(self.current_block_id),
-            self.current_row_id as usize,
-        )
+        // indicate that a new block (located by `current_block_id`) should be
+        // loaded at the beginning of the next `next_batch`
+        self.block_changed = true;
```
skyzh, line 196: If `skip` is called consequently for two times, things would go wrong. On L168, we will never be able to get the correct remaining_items from the block. `let remaining_items = self.block_iterator.remaining_items();`

Personally, I would still prefer retaining the `FakeBlockIterator`, so that the logic could be simple for all block kinds.

skyzh: Please also add unit tests for the `skip` function?

skyzh (commented):
Rest LGTM, good work!

src/storage/secondary/column/concrete_column_iterator.rs
zzl200012: updated

skyzh (approved):
LGTM, but there is a small optimization we can do. If we are already using a fake iterator, it would be better not to use `self.block_iterator` any more.

https://github.com/risinglightdb/risinglight/blob/cb5a0bbf14fa1596925063cb85e3b95e768e6449/src/storage/secondary/column/concrete_column_iterator.rs#L172-L182

It's better not to access `self.block_iterator` any more when `is_fake_iter = true`.

Good work, and thanks for your contribution!

zzl200012:
> LGTM, but there is a small optimization we can do. If we are already using a fake iterator, it would be better not to use `self.block_iterator` any more.
> 
> https://github.com/risinglightdb/risinglight/blob/cb5a0bbf14fa1596925063cb85e3b95e768e6449/src/storage/secondary/column/concrete_column_iterator.rs#L172-L182
> 
> It's better not to access `self.block_iterator` any more when `is_fake_iter = true`.
> 
> Good work, and thanks for your contribution!

Make sense. I'll do this later.

PR #466, opened 2022-02-14
feat(array): introduce DataChunkBuilder and split chunks in HashJoinExecutor

TennyZhuang:
The PR introduces `DataChunkBuilder`, which is responsible for two things:

1. Ensure that all columns are pushed with the same size.
2. Ensure that the size of the DataChunk should not be greater than `capacity`.


wangrunji0408 (commented):
I have two suggestions:
1. Don't consume the builder when the size reaches the limit, so that users will no longer need to rebuild it by themselves.
```rust
impl DataChunkBuilder {
    #[must_use = "should yield the DataChunk when size limit is reached"]
    pub fn push_row(&mut self, row: impl IntoIterator<Item = DataValue>) -> Option<DataChunk>;

    /// Generate a [`DataChunk`] with the remaining rows, leaving the builder empty.
    #[must_use]
    pub fn take(&mut self) -> Option<DataChunk>;
}
```
2. Provide a helper macro to simplify the usage.
```rust
macro_rules! yield_if_some {
    ($maybe_chunk:expr) => {
        if let Some(chunk) = $maybe_chunk {
            yield chunk;
        }
    }
}
```

Then the usage should be something like:
```rust
let mut builder = DataChunkBuilder::new(...);
loop {
    // ...
    yield_if_some!(builder.push_row(values));
    // if necessary
    yield_if_some!(builder.take());
}
yield_if_some!(builder.take());
```

TennyZhuang:
> Don't consume the builder when the size reaches the limit, so that users will no longer need to rebuild it by themselves.

But it can force users to handle the finished DataChunk correctly by compiler error instead of just lint (

TennyZhuang:
> Provide a helper macro to simplify the usage.

Unfortunately, macros does not work well with `yield`.

src/array/data_chunk_builder.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /// Push a row in the Iterator.
+    ///
+    /// The row is accepted as an iterator of [`DataValue`], and it's required that the size of row
+    /// should be the same as the number of columns.
+    ///
+    /// Either a [`DataChunkBuilder`] or a [`DataChunk`] will be returned, depends on whether `size
+    /// == capacity`
+    pub fn push_row(mut self, row: impl IntoIterator<Item = DataValue>) -> Result<Self, DataChunk> {
```
liurenjie1024, line 35: What if we need to append multi rows? Call this one by one may introduce many unnecessary mathod calls.

TennyZhuang: Of course, we can introduce push_rows on demand, but it's not necessary now :(

liurenjie1024: But it may affect signature.

TennyZhuang: It will be another method

st1page: maybe
```rust
    // panic if columns have different size
    pub fn push_multi_rows(mut self, columns: impl IntoIterator<Item = Array>) -> Result<Self, DataChunk> {
```

TennyZhuang:
@wangrunji0408 PTAL

src/array/data_chunk_builder.rs
```diff
@@ -5,12 +5,23 @@ use itertools::Itertools;
...
+/// Panic if there were remaining rows in the builder.
 pub struct DataChunkBuilder {
     array_builders: Vec<ArrayBuilderImpl>,
     size: usize,
     capacity: usize,
 }
 
+impl Drop for DataChunkBuilder {
+    fn drop(&mut self) {
+        assert_eq!(self.size, 0);
```
wangrunji0408, line 21: It's better to give error message here.
```suggestion
        assert_eq!(self.size, 0, "dropping non-empty data chunk builder");
```

src/executor/hash_join.rs
```diff
@@ -106,17 +98,13 @@ impl HashJoinExecutor {
...
-                    }
-                    Ok(builder) => builder,
+                if let Some(chunk) = builder.push_row(values) {
+                    yield chunk;
                 }
             }
         }
 
-        if let Some(chunk) = builder.finish() {
+        if let Some(chunk) = { builder }.take() {
```
wangrunji0408, line 107: Is this `{}` necessary?

src/executor/hash_join.rs
```diff
@@ -38,6 +38,7 @@ impl HashJoinExecutor {
         // helper functions
         let left_rows = || left_chunks.iter().flat_map(|chunk| chunk.rows());
         let right_rows = || right_chunks.iter().flat_map(|chunk| chunk.rows());
+        let data_types = || self.left_types.iter().chain(self.right_types.iter());
```
wangrunji0408, line 41: Now the closure can be removed?
```suggestion
        let data_types = self.left_types.iter().chain(self.right_types.iter());
```

wangrunji0408 (commented):
LGTM in general. Sorry for the late response ==

src/executor/hash_join.rs
TennyZhuang: Not necessary, but consume it explicitly is also reasonable.

wangrunji0408 (approved):

Issue #423, opened 2022-02-07
cli: support graceful stop of txn


skyzh:
Currently, we cannot cancel a query due to:

* We spawn futures over the runtime, and some channels will close after the query is terminated. This would cause problems for channel send.
* When we drop a future, we will drop the transaction object without aborting or committing.

Therefore, we should refactor parts of RisingLight in order to support graceful stop of a transaction.

arkbriar:
After #467 's merged, I think the current executors works properly on interruption.

Shall we close this issue?

PR #467, opened 2022-02-14
refactor(executor): experimental support of cancellable query execution.

arkbriar:
I'm going to close #422 because it's far more complicated to interrupt a query and make things right when a query is cancelled, such as aborting/rolling back transactions and cleaning up spawned tasks.

This PR is not only to #423. Here're the main contributions of this PR:

1. Use cancellation token to signal and detect the cancellation.
2. Provides a non-invasive interface `cancellable` to provide the cancellation of executors and integrates it inside the executor tree. 
  + IMO, only the leaf nodes with much IO works should be sensitive to the cancellation.
  + Resources such as transactions are rollbacked in a spawned task. We think spawned tasks as completable tasks as long as the async runtime is still running. This avoids the complex vacuum thread and cleans the room ASAP.
3. Provide a struct and a set of interfaces to organize the spawned tasks during execution, which avoids leaks while shutting down the async runtime (exit process immediately after interrupting the query).

Here I only include the cancellation of `TableScanExecutor` in this PR. However, the others (e.g. insert/delete/copy) can be supported in the same fashion.

Now you can interrupt a select query in the CLI at any time as you type the CTRL-C. The transactions shall be dealt properly.

To show the debug messages below, please run with `RUST_LOG=risinglight=debug cargo run --release`.

```bash
RUST_LOG=risinglight=debug cargo run --release
> select
    l_orderkey,
    sum(l_extendedprice * (1 - l_discount)) as revenue,
    o_orderdate,
    o_shippriority
from
    customer,
    orders,
    lineitem
where
    c_mktsegment = 'BUILDING'
    and c_custkey = o_custkey
    and l_orderkey = o_orderkey
    and o_orderdate < date '1995-03-15'
    and l_shipdate > date '1995-03-15'
group by
    l_orderkey,
    o_orderdate,
    o_shippriority
order by
    revenue desc,
    o_orderdate
limit 10;

...
^CInterrupted
2022-02-14T15:15:20.381454Z  WARN risinglight::executor::table_scan: Abort!
2022-02-14T15:15:20.381506Z DEBUG risinglight::db: error: abort
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ sqlparser = { git = "https://github.com/risinglightdb/sqlparser-rs", rev = "edea
 thiserror = "1"
 tikv-jemallocator = { version = "0.4", optional = true }
 tokio = { version = "1", features = ["full"] }
-tokio-util = "0.7.0"
+tokio-util = "0.7"
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter", "parking_lot"] }
 manifest-dir-macros = "0.1.11"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ sqlparser = { git = "https://github.com/risinglightdb/sqlparser-rs", rev = "edea
 thiserror = "1"
 tikv-jemallocator = { version = "0.4", optional = true }
 tokio = { version = "1", features = ["full"] }
+tokio-util = "0.7.0"
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter", "parking_lot"] }
 manifest-dir-macros = "0.1.11"
--- a/src/executor/mod.rs
+++ b/src/executor/mod.rs
@@ -12,11 +12,13 @@
 //!
 //! [`try_stream`]: async_stream::try_stream
 
+use std::future::Future;
 use std::sync::Arc;
 
 use futures::stream::{BoxStream, StreamExt};
 use futures_async_stream::try_stream;
 use itertools::Itertools;
+use tokio_util::sync::CancellationToken;
 
 use crate::array::DataChunk;
 use crate::binder::BoundExpr;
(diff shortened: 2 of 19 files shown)


Cargo.toml
```diff
@@ -50,6 +50,7 @@ sqlparser = { git = "https://github.com/risinglightdb/sqlparser-rs", rev = "edea
 thiserror = "1"
 tikv-jemallocator = { version = "0.4", optional = true }
 tokio = { version = "1", features = ["full"] }
+tokio-util = "0.7.0"
```
TennyZhuang, line 53:
```suggestion
tokio-util = "0.7"
```

src/executor/mod.rs
```diff
@@ -114,20 +120,89 @@ pub type BoxedExecutor = BoxStream<'static, Result<DataChunk, ExecutorError>>;
 /// The builder of executor.
 #[derive(Clone)]
 pub struct ExecutorBuilder {
+    context: Arc<Context>,
     storage: StorageImpl,
 }
 
 impl ExecutorBuilder {
     /// Create a new executor builder.
     pub fn new(storage: StorageImpl) -> ExecutorBuilder {
```
TennyZhuang, line 129: Do we even need the method now?

arkbriar: Not really, just for backward compatibility. Remove it should be nicer.

TennyZhuang (approved):
Great work! Generally LGTM, but it seems that we may need some more comments.

TennyZhuang:
And please resolve conflicts.

arkbriar:
> Great work! Generally LGTM, but it seems that we may need some more comments.

Thanks @TennyZhuang . I am greatly encouraged.

I added more commends/docs on `SealableAtomicCounter` and `WaitGroup` and created a new package called `utils` and moved the `sync` package to be under the `utils`. 

Also, 4 more executors (insert/delete, copy from/to file) are refactored and they are all cancellable now. The leaf executors are all cancellable and I think we can interrupt any query from now on.

> And please resolve conflicts.

All conflicts should be resolved now.

Please review this again when you have time.

TennyZhuang:
Thanks! I think it’s good enough to merge!

Issue #456, opened 2022-02-13
cli: output table is splitted


xxchan:
not elegant
```
> create table x (x int)
created
> insert into x values (1)
1 rows inserted
> insert into x values (2)
1 rows inserted
> insert into x values (33333)
1 rows inserted
> select * from x;
+---+
| x |
+---+
| 1 |
+---+
+---+
| 2 |
+---+
+-------+
| 33333 |
+-------+
```

skyzh:
Does this happen in in-memory engine or secondary engine (or both)?

xxchan:
Only in-memory engine.

xxchan:
The table message is implemented in `Display` for `DataChunk`. But the output is `Vec<DataChunk>`. And currently we simply iterate the vec and print each chunk. So if there are multiple chunks in the result, the table will be splitted.

```rust
// in main.rs
for chunk in chunks {
    print_chunk(&chunk)
}

// in db.rs
output[0].set_header(column_names);
```

skyzh:
It seems that we should merge chunks together in `main.rs` before printing them out... As we don't expect users to use cli to print large amount of information, I think it's reasonable to do the fix.

xiaoyong-z:
i will try this

Issue #442, opened 2022-02-11
optimizer: move chunk header derivation to optimizer


skyzh:
In https://github.com/risinglightdb/risinglight/commit/38482e05b1a89690d610c8aeddaa2f6be13b18b1, we added insert / create column name by simply setting the data chunk header. In the future, the column name should always be derived in planner (optimizer).

PR #479, opened 2022-02-16
fix(cli): output-table-is-splitted

xiaoyong-z:

Issue: #456

If the datachunk is more than 1, then stack [1,...] chunks to chunk 0


src/db.rs
```diff
@@ -198,6 +198,10 @@ impl Database {
                 debug!("output:\n{}", chunk);
             }
             if !column_names.is_empty() && !output.is_empty() {
+                if output.len() > 1 {
+                    let chunks: Vec<DataChunk> = output.drain(1..).collect();
+                    output[0].stack(chunks);
+                }
```
xxchan, line 204: Would it be better to merge chunks in `main.rs` (e.g., change `print_chunk(chunk: &DataChunk)` to print chunks) instead of `run`? `run` returns `Vec<DataChunk>`. We'd better not change it?

src/array/data_chunk.rs
```diff
@@ -129,6 +129,23 @@ impl DataChunk {
         }
         arrays.into_iter().collect()
     }
+
+    pub fn stack(&mut self, chunks: Vec<data_chunk::DataChunk>) {
```
xxchan, line 133:
```suggestion
    pub fn stack(&mut self, chunks: &Vec<Self>) {
```

src/array/data_chunk.rs
```diff
@@ -129,6 +129,23 @@ impl DataChunk {
...
+        for col_idx in 0..self.column_count() {
+            let mut builder = ArrayBuilderImpl::from_type_of_array(self.array_at(col_idx));
+            for row in self.rows() {
+                builder.push(&row.get(col_idx))
+            }
+            for chunk in &chunks {
+                for row in chunk.rows() {
+                    builder.push(&row.get(col_idx))
+                }
+            }
```
xxchan, line 144: `ArrayBuilder` can `append` an array directly. We don't need to `push` row by row

src/db.rs
xiaoyong-z: User may submit multiple statments to client at once, 
For example, the input is
```sql
create table t; insert into x values (1); insert into x values (2); select * from x; insert into x values (3333);
```
then there might be multiple statements in a `run`, how to handle this case in the print_chunks? I think we need to add some complex logic.

xxchan: Oh this is a good catch! So each element of the returned `Vec` of `run` is the results for one command. I think it makes sense.

src/db.rs
```diff
@@ -198,6 +198,10 @@ impl Database {
                 debug!("output:\n{}", chunk);
             }
             if !column_names.is_empty() && !output.is_empty() {
+                if output.len() > 1 {
+                    let chunks: Vec<DataChunk> = output.drain(1..).collect();
+                    output[0].stack(&chunks);
+                }
                 output[0].set_header(column_names);
             }
             outputs.extend(output);
```
xxchan, line 207: We can replace `extend` by `push` now to make it more clear?

src/db.rs
```diff
@@ -198,6 +198,10 @@ impl Database {
                 debug!("output:\n{}", chunk);
             }
             if !column_names.is_empty() && !output.is_empty() {
+                if output.len() > 1 {
+                    let chunks: Vec<DataChunk> = output.drain(1..).collect();
+                    output[0].stack(&chunks);
```
xxchan, line 203: We can let `stack` accpets an iterator so the `collect` is not needed?

xxchan (approved):
Others LGTM. Thanks for your contribution!

src/db.rs
xiaoyong-z: well，i am not very familar with rust. 
I try this one first: `outputs.push(output[0])`,  and the complier warns me the DataChunk needs to implement the copy trait, i think copy the output[0] to outputs cost a lot. 
So here i replace it by `outputs.push(output.remove(0))` to move the output[0] to outputs. However, i think it is not elegant. 
Can you give me more suggestions? Any suggestions are welcome.

src/db.rs
xiaoyong-z: here too, i try the code `output[0].stack(output.drain(1..))` and `output[0].stack(&output.iter())` first, and the complier. warns me that there are mutable borrows and one immutable borrows.
So i first stacks the vector into one temp variaible chunk, and then move the chunk to output[0]

skyzh (approved):
Thanks for your contribution! But I'd prefer to keep this PR on hold.

After the discussion above, I found that we cannot simply merge chunks in `main.rs`, because the chunks might be produced by multiple queries. However, if we add the merge logic in `db.rs`, all queries, including cli queries and embedded ones, will suffer from the performance regression caused by chunk merging.

The merging logic is only needed in cli -- when we print chunk to the console, the `comfy-table` need to calculate width in advance, so we have to feed a whole chunk to that.

Therefore, I have some further ideas to solve this issue... Any of them might work.

* The `run` function should return `Vec<Vec<DataChunk>>` (... and `Vec<(Header, Vec<DataChunk>)>` in the future when we have finished the column name deduction framework.) Each of the element represents the result of one query. Then, we can freely merge chunks in `main.rs` or not merge them when running in embedded mode.
* Modify `comfy-table` to first calculate width in advance based on multiple chunks, and print them one by one.
* If there are >=2 chunks from `run`, print them in plain-text format instead of table format.

src/db.rs
xxchan: oh I understand your problem. I think there are several ways

First your current solution looks good to me. But I think we can use the temp variable directly without moveing it into output[0]?

Alternately we can also write things like 
```rust 
let it=output.into_iter();
let chunk=it.next().unwrap();
chunk.stack(it)
```

```

File: src/array/data_chunk.rs [hunk 6]
```diff
@@ -129,9 +129,25 @@ impl DataChunk {
...
+        for col_idx in 0..first_chunk.column_count() {
+            let mut builder = ArrayBuilderImpl::from_type_of_array(first_chunk.array_at(col_idx));
+            for chunk in chunks {
+                builder.append(chunk.array_at(col_idx));
+            }
+            arrays.push(builder.finish());
+        }
+        DataChunk {
+            arrays: arrays.into_iter().collect(),
+            header: None,
```
skyzh, line 145: Should preserve the header from the first chunk?

skyzh (approved):
LGTM, thanks!

The logic around `header` looks tedious. Considering we will eventually remove data chunk header https://github.com/risinglightdb/risinglight/issues/442, I'll give this PR an approval.

Thanks!

skyzh:
Oops, you'll need to resolve conflicts :(

src/array/data_chunk.rs
xiaoyong-z: this function is unused now , i forget to delete this function

PR #481, opened 2022-02-16
fix(cli): don't panic on bind error

### LINKED ISSUE #448: cli: don't panic on bind error.
image

we can hint `parse error:xxx` like other databases.

xiaoyong-z:

close #448 

1. warning users for show xxx
2. binder not panic


src/db.rs
```diff
@@ -170,6 +170,12 @@ impl Database {
         // parse
         let stmts = parse(sql)?;
 
+        if sql.to_lowercase().starts_with("show ") {
```
skyzh, line 173: I would suggest returning error in the binder level, instead of simply doing string matching. What if user input is `<space>show tables`?

You may add a new match arm at 

https://github.com/risinglightdb/risinglight/blob/de55dfe2b53e9d6d623e330d5f15d40e721cfa8a/src/binder/mod.rs#L110-L122

And return a special error (like SQL not supported) for `show` statements.

xiaoyong-z:
I update the commit. But i still have a question:
As you can see in the below, there are three types of show statements, "show tables" and "show databases" belongs to the ShowVariable type. However, the comment says that it is a 
PostgreSQL-specific statement, which means that pg also need to support show tables? 
Btw, the code is from the sqlparser-rs
```rust
    /// Note: this is a PostgreSQL-specific statement.
    ShowVariable { variable: Vec<Ident> },
    /// SHOW CREATE TABLE
    ///
    /// Note: this is a MySQL-specific statement.
    ShowCreate {
        obj_type: ShowCreateObject,
        obj_name: ObjectName,
    },
    /// SHOW COLUMNS
    ///
    /// Note: this is a MySQL-specific statement.
    ShowColumns {
        extended: bool,
        full: bool,
        table_name: ObjectName,
        filter: Option<ShowStatementFilter>,
    },
```

skyzh:
sqlparser-rs is a general-purpose parser... And I think we should take `psql` as reference instead of sqlparser-rs.

skyzh (approved):
LGTM, thanks!

Issue #102, opened 2021-11-03
binder: `insert into .. select .. from ..` support


skyzh:
This is useful for batch writes and data import. We can later simply:

```
INSERT INTO my_table SELECT * FROM 'table.tbl'
```

MingjiHan99:
I could have a try

PR #487, opened 2022-02-16
feat(insert): support `insert into select from`

MingjiHan99:
close https://github.com/risinglightdb/risinglight/issues/102


skyzh (approved):
Good work!

@st1page said that we can add a test for `insert into t select * from t` 🤪

PR #489, opened 2022-02-17
fix(storage): fix panic when finishing an empty blob builder.

### LINKED ISSUE #463: enable blob test for disk
Discussion from https://github.com/risinglightdb/risinglight/pull/461

Currently the test failed because
```
thread 'tokio-runtime-worker' panicked at 'no user column', src/storage/secondary/rowset/rowset_iterator.rs:56:13
```
This is caused by `DELETE FROM blobs WHERE true`, which should be fixed in https://github.com/risinglightdb/risinglight/issues/421. We can turn it on afther the issue fixed.

arkbriar:

Fixes #463 .

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -10,7 +10,7 @@ fn main() {
     println!("cargo:rerun-if-changed=tests/sql");
     const PATTERN: &str = "tests/sql/**/[!_]*.slt"; // ignore files start with '_'
     const MEM_BLOCKLIST: &[&str] = &["statistics.slt"];
-    const DISK_BLOCKLIST: &[&str] = &["blob.slt"];
+    const DISK_BLOCKLIST: &[&str] = &[];
 
     let path = PathBuf::from(env::var("OUT_DIR").unwrap()).join("testcase.rs");
     let mut fout = std::fs::File::create(path).expect("failed to create file");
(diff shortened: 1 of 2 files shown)


skyzh (commented):
If there are no data in txn, the RowSetBuilder's finish should never be called.

... I don't think this is the root cause of the problem. Maybe we should avoid `RowSetBuilder::finish_and_flush` to be called if there are no new rows in a txn?

arkbriar:
> If there are no data in txn, the RowSetBuilder's finish should never be called.
> 
> ... I don't think this is the root cause of the problem. Maybe we should avoid `RowSetBuilder::finish_and_flush` to be called if there are no new rows in a txn?

That's true. But neither the column builder nor the rowset builder rejects an empty `finish` or `flush`.

BTW, I think it's better to separate the flush from the builder so we can control the flush in `SecondaryMemRowset::flush`.

skyzh (approved):
Considering this is a quick fix for https://github.com/risinglightdb/risinglight/issues/463, I'd approve it. But I don't think this is a fix that solves all problems for us -- I suspect the issue will still occur if we `select v from table where true` on a Decimal, or String column.

> BTW, I think it's better to separate the flush from the builder so we can control the flush in SecondaryMemRowset::flush.

Good idea! You may elaborate it in more detail in an issue, or simply propose another fix. Thanks!

Issue #110, opened 2021-11-05
storage: remove I/O operation in `RowSetBuilder`


skyzh:
... and we need a new `RowSetWriter` to write what's inside builder to the disk.

skyzh:
From my perspective, the `RowsetBuilder` should not contain any I/O related logic, and the logic like creating directory and flushing data should be moved to a separate module, maybe `RowsetWriter`? So that we won't need to create a directory even the rowset is still in memory.

_Originally posted by @zzl200012 in https://github.com/risinglightdb/risinglight/pull/492#discussion_r809098455_

> From my perspective, the `RowsetBuilder` should not contain any I/O related logic, and the logic like creating directory and flushing data should be moved to a separate module, maybe `RowsetWriter`? So that we won't need to create a directory even the rowset is still in memory.

That's something I'd like to refactor :)

In fact, all I/O is done in `flush_and_finish` for RowSetBuilder. I plan to let the caller to decide whether to write these files or not, instead of doing all I/O inside the `RowSetBuilder`.

We already have a tracking issue for that https://github.com/risinglightdb/risinglight/issues/110

_Originally posted by @skyzh in https://github.com/risinglightdb/risinglight/pull/492#discussion_r809157267_

zzl200012:
That's fine. I will take a look this weekend.

skyzh:
Oops, it seems that https://github.com/risinglightdb/risinglight/pull/499/ has already fixed this by introducing a RowSet writer. cc @zzl200012

PR #492, opened 2022-02-17
fix(storage): skip flush in rowset builder if it's empty, allow empty `finish` of column builders.

arkbriar:

As mentioned in #489 , the panic is caused by invoking `finish_and_flush` on an empty rowset builder, then finishes an empty column builder. Thanks @skyzh for pointing out my problem.

This PR fixes these problems by allowing finishing empty column builder and skip the `finish_and_flush` of rowset builder entirely if empty.

### DIFF
--- a/src/storage/secondary/rowset/rowset_builder.rs
+++ b/src/storage/secondary/rowset/rowset_builder.rs
@@ -99,6 +99,11 @@ impl RowsetBuilder {
     }
 
     pub async fn finish_and_flush(self) -> StorageResult<()> {
+        // Skip flush when empty.
+        if self.row_cnt == 0 {
+            return Ok(());
+        }
+
         for (column_info, builder) in self.columns.iter().zip(self.builders) {
             let (index, data) = builder.finish();
 
--- a/src/storage/secondary/column/char_column_builder.rs
+++ b/src/storage/secondary/column/char_column_builder.rs
@@ -46,6 +46,10 @@ impl CharColumnBuilder {
     }
 
     fn finish_builder(&mut self) {
+        if self.current_builder.is_none() {
+            return;
+        }
+
         let (block_type, stats, mut block_data) = match self.current_builder.take().unwrap() {
             CharBlockBuilderImpl::PlainFixedChar(builder) => (
                 BlockType::PlainFixedChar,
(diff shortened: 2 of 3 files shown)


src/storage/secondary/rowset/rowset_builder.rs
```diff
@@ -99,6 +99,11 @@ impl RowsetBuilder {
     }
 
     pub async fn finish_and_flush(self) -> StorageResult<()> {
+        // Skip flush when empty.
+        if self.row_cnt == 0 {
+            return Ok(());
```
skyzh, line 104: I would prefer returning an error here. The caller might expect RowSetBuilder to create a directory for the rowset. But it turns out not to be existing.

By returning an error, we can see where we flushed an empty rowset, and therefore fix the upper logic.

arkbriar: That's my fault... I'll just return an error here.

One possible loc is in the compactor. I'll try to reproduce.

zzl200012: Agree with that we should return an error here to notify upper logic. Just a reminder: the rowset directory was created when memory table was created, which was refactored previously. So the only difference here is the empty rowset doesn't have data, rather than directory not exist.

zzl200012: From my perspective, the `RowsetBuilder` should not contain any I/O related logic, and the logic like creating directory and flushing data should be moved to a separate module, maybe `RowsetWriter`? So that we won't need to create a directory even the rowset is still in memory.

arkbriar: I make it panic here. No suitable error is found. I'll go deep into the original test and find out if there's something I've missed.

skyzh: > From my perspective, the `RowsetBuilder` should not contain any I/O related logic, and the logic like creating directory and flushing data should be moved to a separate module, maybe `RowsetWriter`? So that we won't need to create a directory even the rowset is still in memory.

That's something I'd like to refactor :)

In fact, all I/O is done in `flush_and_finish` for RowSetBuilder. I plan to let the caller to decide whether to write these files or not, instead of doing all I/O inside the `RowSetBuilder`.

We already have a tracking issue for that https://github.com/risinglightdb/risinglight/issues/110

skyzh (approved):

PR #497, opened 2022-02-18
fix(storage): ignore errors on shutdown.

arkbriar:

Shutdown of secondary storage might panic when the whole runtime's shutting down. 

It's ok to ignore the errors here.

```bash
thread 'tokio-runtime-worker' panicked at 'empty rowset', src/storage/secondary/rowset/rowset_builder.rs:104:13
thread 'disk::_blob_slt_' panicked at 'called `Result::unwrap()` on an `Err` value: ()', .../risinglight/src/storage/secondary/mod.rs:142:44
```


skyzh (approved):
LGTM. But the error in the PR body `empty rowset` should be fixed...

Issue #256, opened 2021-12-22
Bug(Binder): wrong fields in query with group by.


st1page:
```SQL
> create table t(id int, v1 int, v2 int, v3 int);
> explain select * from t group by id;
+----------------------------------------------------------------------------------------------+
| PhysicalProjection: exprs [InputRef #0, InputRef #1, InputRef #2, InputRef #3]               |
|   PhysicalSeqScan: table #8, columns [0, 1, 2, 3], with_row_handler: false, is_sorted: false |
+----------------------------------------------------------------------------------------------+
> explain select id from t group by id;
+-------------------------------------------------------------------------------------+
| PhysicalProjection: exprs [InputRef #0]                                             |
|   PhysicalSeqScan: table #8, columns [0], with_row_handler: false, is_sorted: false |
+-------------------------------------------------------------------------------------+
```

skyzh:
Is `explain select * from t group by id;` a valid SQL 🤪? We should reject such SQL in our system... We can never know which v1, v2, v3 to select from the group :(

skyzh:
IIRC, SQLite might accept such SQL and only return the first tuple of `v1, v2, v3`.

st1page:
> Is `explain select * from t group by id;` a valid SQL zany_face? We should reject such SQL in our system... We can never know which v1, v2, v3 to select from the group :(

in PG, there will be an error 
> column "t.v1" must appear in the GROUP BY clause or be used in an aggregate function

st1page:
so there are two issues.
1. treat the star (*) in a more general way in the binder, for example expand all columns appears in the context. same issue in #388
2. fix the wrong plan for query `select id from t group by id;` it not recognize the aggregation query correctly and does not give a Agg plan at all.

PR #500, opened 2022-02-18
fix(planner): group by without aggr calls

xxchan:
try to fix https://github.com/risinglightdb/risinglight/issues/256

current result:
```sql
> create table t(id int, v1 int, v2 int, v3 int);
created]
> explain select id from t group by id;
  PhysicalHashAgg: 0 agg calls
    PhysicalTableScan: table #0, columns [0], with_row_handler: false, is_sorted: false, expr: None
> explain select * from t group by id;
thread 'tokio-runtime-worker' panicked at 'column reference not found', src/optimizer/logical_plan_rewriter/input_ref_resolver.rs:30:26
thread 'main' panicked at 'failed to join query thread: JoinError::Panic(...)', src/main.rs:79:23
```


tests/sql/select.slt
```diff
@@ -43,6 +75,19 @@ select count(v3), v2 from t group by v2
 2 3
 2 4
 
+# AggregateWithProjectionTest2
+query II rowsort
+select v2 from t group by v2
+----
+3
+4
```
xxchan, line 83: This is new test. Others are merged from `select.test`

skyzh (approved):
LGTM!

```
> explain select * from t group by id;
thread 'tokio-runtime-worker' panicked at 'column reference not found', src/optimizer/logical_plan_rewriter/input_ref_resolver.rs:30:26
thread 'main' panicked at 'failed to join query thread: JoinError::Panic(...)', src/main.rs:79:23
```

For this panic, we might need to create a new issue for that. Better to reject such query as we didn't apply aggregation functions on columns other than group keys.

Issue #501, opened 2022-02-19
planner: row handler not included when count(*) is used as subexpression


xxchan:
```rust
> explain select count(*)*2 from t
PhysicalProjection: exprs [Multiply(InputRef #0, Int32(2) (const))]
  PhysicalSimpleAgg: 1 agg calls
    PhysicalTableScan: table #10, columns [], with_row_handler: false, is_sorted: false, expr: None

in 2ms 169us 23ns
>  select count(*)*2 from t
thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 0 but the index is 0', src/array/data_chunk.rs:71:10
```

PR #502, opened 2022-02-19
fix(planner): row handler not included when count(*) is used as subexpression

xxchan:
fix https://github.com/risinglightdb/risinglight/issues/501

```sql
> explain select count(*) from t;
  PhysicalSimpleAgg: 1 agg calls
    PhysicalTableScan: table #10, columns [], with_row_handler: true, is_sorted: false, expr: None

> explain select 1 from t;
PhysicalProjection: exprs [Int32(1) (const)]
  PhysicalTableScan: table #10, columns [], with_row_handler: false, is_sorted: false, expr: None

> explain select count(*)+1 from t;
PhysicalProjection: exprs [Plus(InputRef #0, Int32(1) (const))]
  PhysicalSimpleAgg: 1 agg calls
    PhysicalTableScan: table #10, columns [], with_row_handler: true, is_sorted: false, expr: None

> explain select -count(*) from t;
PhysicalProjection: exprs [BoundUnaryOp { op: Minus, expr: InputRef #0, return_type: Some(Int(None)) }]
  PhysicalSimpleAgg: 1 agg calls
    PhysicalTableScan: table #10, columns [], with_row_handler: true, is_sorted: false, expr: None

```

### DIFF
--- a/src/binder/expression/mod.rs
+++ b/src/binder/expression/mod.rs
@@ -92,6 +92,40 @@ impl BoundExpr {
         self.get_filter_column_inner(&mut filter_column);
         filter_column
     }
+
+    pub fn contains_column_ref(&self) -> bool {
+        match self {
+            BoundExpr::Constant(_) => false,
+            BoundExpr::ColumnRef(_) => true,
+            BoundExpr::InputRef(_) => false,
+            BoundExpr::BinaryOp(e) => {
+                e.left_expr.contains_column_ref() || e.right_expr.contains_column_ref()
+            }
+            BoundExpr::UnaryOp(e) => e.expr.contains_column_ref(),
+            BoundExpr::TypeCast(e) => e.expr.contains_column_ref(),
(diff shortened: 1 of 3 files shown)


xxchan:
```sql
> explain select count(*)+sum(v) from t
PhysicalProjection: exprs [Plus(InputRef #0, InputRef #1)]
  PhysicalSimpleAgg: 2 agg calls
    PhysicalTableScan: table #10, columns [0], with_row_handler: false, is_sorted: false, expr: None
```

skyzh (approved):
Thanks!

Would you please add the cases in issue body to sqllogictests?

By the way, I'm thinking of moving row-handler related logics to the optimizer... For example, if column pruning found that there's no column to scan, we automatically attach a row handler to the table.

At the same time, if our queries become more and more complex, it won't be possible for us to enumerate all rules to add row_handler column, solely in the planner... I don't have a clear idea about that, but I think we will eventually (1) make row_handler a real logical column instead of a flag for table (2) use the optimizer rule to decide whether to scan the row handler column.

PR #504, opened 2022-02-19
fix(storage): filter columns will be skipped incorrectly

skyzh:

If `filter_map` is all false when filter condition gets evaluated, as we have already scanned filter columns, we should not skip them.


zzl200012 (approved):
LGTM

Issue #264, opened 2021-12-23
executor: LimitExecutor will panic if all chunks only have 1 element


skyzh:
After https://github.com/singularity-data/risinglight/pull/263 is applied, and `next_batch` is called with:

```rust
it.next_batch(Some(1)).await
```

Limit executor panics with

```
thread 'test_disk::_limit_slt_' panicked at 'attempt to subtract with overflow', src/executor/limit.rs:13:9
```

xxchan:
Is this fixed? I'm a little bit confused about how to reproduce this issue. Where to call `it.next_batch(Some(1)).await`?

skyzh:
My fault. There are two ways to reproduce this:

* Change `TableScanExecutor` to return one row at a time.

https://github.com/risinglightdb/risinglight/blob/caa809667b5ccf41cceef17a6ac54fb35369680c/src/executor/table_scan.rs#L110

where we can replace `next_batch(None)` with `next_batch(Some(1))`.

```
---- disk::_limit_slt_ stdout ----
thread 'disk::_limit_slt_' panicked at 'attempt to subtract with overflow', src/executor/limit.rs:28:23
```

and meanwhile, I found that a lot of other tests failed if we return one row at a time 🤣

* I guess if we simply insert multiple one-row chunk to in-memory storage, and `select * from table limit 10`, will also reproduce this issue.

skyzh:
When batch size == 1, all tests other than `limit` has been fixed in https://github.com/risinglightdb/risinglight/pull/504/

PR #506, opened 2022-02-19
fix(executor): limit should exit early when offset and limit are both 0

xxchan:
Fix https://github.com/risinglightdb/risinglight/issues/264

The issue can be reproduced using memory engine like this: (Btw, the sqlogictest fails for both disk & memory engines)
```sql
> create table t(x int);insert into t values (0);insert into t values (0);
> select * from t limit 0;
```
The execution process is 
```
processed 0, cardinality 1, offsef 0, limit 0
processed 1, cardinality 1, offsef 0, limit 0 # boom
```

Because when `offset=limit=0`, `start=end=0`, and thus it will `continue`.
https://github.com/risinglightdb/risinglight/blob/fea5e0b3de264e3f2275d1ca6c04c2c97ef8d639/src/executor/limit.rs#L27-L32

### DIFF
--- a/src/executor/limit.rs
+++ b/src/executor/limit.rs
@@ -22,6 +22,9 @@ impl LimitExecutor {
             let batch = batch?;
             if dummy_chunk.is_none() {
                 dummy_chunk = Some(batch.slice(0..0));
+                if self.offset == 0 && self.limit == 0 {
+                    break;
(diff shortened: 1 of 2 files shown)


skyzh (approved):
Thanks!

PR #507, opened 2022-02-20
feat(storage): support run-length encoding

### LINKED ISSUE #255: storage: support run-length encoding
It is a common case that users are storing nearly the same value in one column. For example, in TPC-H, we have a `CHAR(1)` column. There must be duplicated data in this column if we have 1000k rows, and therefore we can use running-length encoding to do some compression.

# Guide-level Introduction

RLE encoding is a compression scheme instead of a typical encoding. It can be applied on any column. For example, given a int column:

```
1, 1, 2, 2, 3, 3
-> PrimitiveBlockBuilder(1, 1, 2, 2, 3, 3)
```

RLE encoding will check if the current value is the same as the previous one, and generate a RLE map:

```
1, 1, 2, 2, 3, 3
-> PrimitiveBlockBuilder(1, 2, 3)
-> RLE(2, 2, 2)
```

# Implementation-level Introduction

## Builder and Iterator

Contrary to primitive iterators, RLE block builder and iterator are simply a wrapper on primitive iterators. It will look like follows:

```
pub struct RLEBlockBuilder<B: BlockBuilder> {
  block_builder: B,
  rle_count: Vec<u16>,
  previous_value: Option<B::ArrayType::OwnedItem>
}
```

When finishing a block, RLEBlockBuilder simply append the RLE count data to the encoded block.

## Column

L0 rowsets should not use RLE encoding. We can use distinct value statistics to decide whether to use RLE encoding to the compacted new rowsets.

We should modify factories in column correspondingly:

```rust
pub(super) enum BlockIteratorImpl<T: PrimitiveFixedWidthEncode> {
    Plain(PlainPrimitiveBlockIterator<T>),
    PlainNullable(PlainPrimitiveNullableBlockIterator<T>),
    /* new */ RLEPlainNullable(RLEBlockIterator<PlainPrimitiveNullableBlockIterator<T>>)
}
```

And in proto:

```
  enum BlockType {
    Plain = 0;
    RunLength = 1;
    ZstdCompress = 2;
    PlainNullable = 3;
    PlainFixedChar = 4;
    PlainVarchar = 5;
    RLEPlainNullable = 6; /* new */
  }
```

ludics:
storage: support run-length encoding #255

- add 3 BlockType with prefix `Rle`
- add `rle_block_builder`, `rle_block_iterator`
- update corresponding `ColumnBuilder` and `ColumnIterator`

close #255

Signed-off-by: Lu Di [leonludics@gmail.com](mailto:leonludics@gmail.com)


yingjunwu:
It's called run-length encoding (RLE), not running-length encoding.

skyzh (commented):
Thanks a lot for this feature! But before proceeding to review, I'd like to know why we need 2 separate RLEBuilder / RLEIterator for variable / fixed-length types? If we write bounds from the `Array`:

```
pub struct RleBlockIterator<A, B>
where
    A: Array,
    B: BlockIterator<A>,
```

This should cover all cases, including both primitive types and Bytes types. Did you meet any difficulty when implementing in this way?

skyzh:
By the way, I've force enabled is_rle and run TPC-H tests on this branch. It seems that...

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TestError { kind: QueryResultMismatch { sql: "select     sum(l_extendedprice * l_discount) as revenue from     lineitem where     l_shipdate >= date '1994-01-01'     and l_shipdate < date '1994-01-01' + interval '1' year     and l_discount between 0.08 - 0.01 and 0.08 + 0.01     and l_quantity < 24;", expected: "163821038.3323", actual: "163823134.1083" }, loc: Location { file: "tests/sql/tpch-full/_q6.slt", line: 1 } }', src/main.rs:217:41
```

Some queries are producing wrong results. See https://github.com/risinglightdb/risinglight/blob/main/docs/01-tpch.md#developers-add-new-tpc-h-tests on how to run TPC-H tests with RisingLight.

Don't worry about that, I think I might find the issue when I review.

proto/src/proto/rowset.proto
```diff
@@ -35,6 +35,9 @@ message BlockIndex {
     PlainNullable = 3;
     PlainFixedChar = 4;
     PlainVarchar = 5;
+    RlePainNullable = 6;
+    RlePlainFixedChar = 7;
+    RlePlainVarchar = 8;
```
xxchan, line 40: I think `Plain` should be removed, since `Rle` and `Plain` are the same level concept? `Plain` for plain text encoding

xxchan:
If there are difficulties in implementing RLE for non-primitive types, I think we'd better not implement it to keep our codebase simple. Because RLE may not be very useful for non-primitive types?

skyzh:
> Because RLE may not be very useful for non-primitive types?

There's a one-char column `l_returnflag` in `lineitem` table in TPC-H test. RLE is still helpful for those using char to store enum-like types.

ludics:
> This should cover all cases, including both primitive types and Bytes types. Did you meet any difficulty when implementing in this way?

In commit [f163ed](https://github.com/risinglightdb/risinglight/pull/507/commits/f163edf0924bf268971a839178819aae20015dc9), I added a new trait RLETypeEncode and implemented only one builder/iterator for both primitive types and Bytes types. However, the `RLEBlockBuilder` is not very efficient as the `previous_value` is stored as `Option<Vec<u8>>` and the item to be appended will also be transferred to `Vec<u8>` for primitive types to make a comparison. Storing `previous_value` as `Option<&[u8]>` is more efficient, but there will be some lifetime problems...

skyzh:
We may store previous value as <A::Item as ToOwned>::Owned? This is equivalent to i32, String, etc.

src/array/mod.rs
```diff
@@ -72,7 +72,7 @@ pub trait Array: Sized + Send + Sync + 'static {
     type Builder: ArrayBuilder<Array = Self>;
 
     /// Type of element in the array.
-    type Item: ToOwned + ?Sized;
+    type Item: ToOwned + ?Sized + PartialEq;
```
skyzh, line 75: I'd prefer write the `PartialEq` bound on where we need, instead of on all items. For example,

```
impl<A, B> RleBlockBuilder<A, B>
 where
     A: Array,
     B: BlockBuilder<A>,
     A::Item: PartialEq,
```

src/array/utf8_array.rs
```diff
@@ -19,7 +19,7 @@ pub struct BytesArray<T: ValueRef + ?Sized> {
 }
 
 /// The borrowed type of a variable-length value.
-pub trait ValueRef: ToOwned + AsRef<[u8]> + Send + Sync + 'static {
+pub trait ValueRef: ToOwned + AsRef<[u8]> + Send + Sync + 'static + PartialEq {
```
skyzh, line 22: ditto, better not to add `PartialEq` bound here.

src/storage/secondary/block.rs
```diff
@@ -53,7 +57,10 @@ pub trait BlockBuilder<A: Array> {
     /// Get estimated size of block. Will be useful on runlength or compression encoding.
     fn estimated_size(&self) -> usize;
 
-    /// Get distinct values count of block
+    /// Get the size to append one data into the block
+    fn size_to_append(&self, item: &Option<&A::Item>) -> usize;
```
skyzh, line 61: It seems that this function is solely used for determining `should_finish`. However, `should_finish` doesn't need to be accurate -- we can finish early, or slightly exceed the block size. Therefore, I suggest removing this function.

src/storage/secondary/block/rle_block_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+            rle_counts: Vec::new(),
+            previous_value: None,
+            target_size,
+        }
+    }
+
+    fn append_inner(&mut self, item: Option<&A::Item>) {
+        self.previous_value = item.map(|x| x.to_owned());
+        self.block_builder.append(item);
+        self.rle_counts.push(1);
```
skyzh, line 44: One optimization is to store the current `rle_count` as a u16 in the struct (instead of as last element in Vec), and push the previous count to the Vec when we append the next datum. Therefore, we can avoid bound check every time we access the last item.

src/storage/secondary/block/rle_block_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    fn should_append_inner(&self, item: &Option<&A::Item>) -> bool {
+        let len = self.rle_counts.len();
+        if len == 0 {
+            return true;
+        }
+        if let Some(item) = item {
```
skyzh, line 52: If T is PartialEq, then Option<T> is also PartialEq. Therefore, we can directly compare:

```
if item == self.previous.value && self.rle_counts.last().unwrap() < u16::MAX {
  // ...
} else {
  // ...
}
```

skyzh: Also we can consider using variable size encoding (IIRC, prost::encode_varint?) for those integers, and therefore we can store elements more efficiently, and we don't need to limit the RLE count to u16. This could be done in the future; I'm not sure if we can optimize storage space by using variable size encoding for RLE count.

src/storage/secondary/block/rle_block_iterator.rs
```diff
@@ -0,0 +1,451 @@
...
+use super::{Block, BlockIterator};
+use crate::array::{Array, ArrayBuilder};
+
+pub fn decode_rle_block(data: Block) -> (usize, Block, Block) {
+    let mut buffer = &data[..];
+    let rle_num = buffer.get_u32_le() as usize;
+    let rle_length = std::mem::size_of::<u32>() + std::mem::size_of::<u16>() * rle_num;
+    let rle_data = data[std::mem::size_of::<u32>()..rle_length].to_vec();
+    let block_data = data[rle_length..].to_vec();
+    (rle_num, Bytes::from(rle_data), Bytes::from(block_data))
```
skyzh, line 14: We can directly use `data.slice(..)`, therefore to reduce copy. By using `.slice`, we can have two objects referring to a single Bytes memory space, and we don't need extra allocation.

See https://docs.rs/bytes/latest/bytes/ for more information.

src/storage/secondary/block/rle_block_iterator.rs
```diff
@@ -0,0 +1,451 @@
...
+                self.cur_row += 1;
+                self.block_iter.skip(1);
+                if self.cur_row >= self.rle_row_count {
+                    break;
+                }
+            }
+        }
+    }
+
+    fn remaining_items(&self) -> usize {
```
skyzh, line 142: This remaining_items implementation is too heavy. Is it possible to calculate the total row in advance (e.g., in `new`), and simply return `total_row - current_row_id`?

src/storage/secondary/column/primitive_column_builder.rs
```diff
@@ -111,11 +124,32 @@ impl<T: PrimitiveFixedWidthEncode> ColumnBuilder<T::ArrayType> for PrimitiveColu
         while iter.peek().is_some() {
             if self.current_builder.is_none() {
                 if self.nullable {
-                    self.current_builder = Some(BlockBuilderImpl::PlainNullable(
-                        PlainPrimitiveNullableBlockBuilder::new(
-                            self.options.target_block_size - 16,
-                        ),
-                    ));
+                    if self.options.is_rle {
```
skyzh, line 127: Consider using `match` here, otherwise it's hard to read...

```rust
match (self.nullable, self.options.is_rle) {
  (true, true) => xxx,
  (true, false) => xxx,
  // ...
}
```

skyzh (commented):
Rest LGTM, excellent work!

src/storage/secondary/block.rs
ludics: So for `should_finish` of `RleBlockBuilder`, will it be ok that only adding `std::mem::size_of::<u16>()` when really append an item for `self.block_builder`?

skyzh: I would prefer first check if the child block builder “should finish”, if not, simply add u16 (for a single RLE count).

ludics: Currently, for `RleBlockBuilder`, we don't care about the target size of child block builder because it's hard to predict when using RLE, and we only care about the total size together with `rle_counts`. If we first check whether the child block builder "should finish", then we need to set a target size for `RleBlockBuilder` and child block builder respectively.

src/storage/secondary/block/rle_block_builder.rs
ludics: > Also we can consider using variable size encoding (IIRC, prost::encode_varint?) for those integers, and therefore we can store elements more efficiently, and we don't need to limit the RLE count to u16. This could be done in the future; I'm not sure if we can optimize storage space by using variable size encoding for RLE count.

Add an issue for this?

src/storage/secondary/block.rs
skyzh: After a second though, I think we should go simple here, just return `self.block_builder.should_finish(item)`.

It is true that this is not accurate -- when there are no items of the same in block, we will produce larger block; when there are a lot of same large items, it will produce smaller block. But this function will return true eventually.

Therefore, we don't need to care about the exact size.

What do you think?

src/storage/secondary/block/rle_block_builder.rs
skyzh: Yes, you may create one when this PR gets merged.

src/storage/secondary/block.rs
skyzh: My point is that, the `append_one_by_one` function will:

1. first check `should_finish`.
2. then append.

If we have some complex operations, like doing comparison in `should_finish`, the comparison will be done twice in (1) and (2). So I just want the `should_finish` function to be as simple as possible...

skyzh (approved):
Rest LGTM! You may get it merged and fix the remaining small problems later. Thanks for your contribution.

I'm still investigating why enabling RLE will produce wrong result in TPC-H. I'll create an issue later about this problem.

skyzh:
Well, I've finally found the issue with `skip`. I'll start another issue on this.

PR #509, opened 2022-02-21
tests: use `AsyncDB` (re #485)

### LINKED ISSUE #485: tests: bump sqllogictest to latest main and use async interface
In the latest unreleased https://github.com/singularity-data/sqllogictest-rs/, we added a `AsyncDB` trait. We should replace the current mock-async implementation with pure async fn.

https://github.com/risinglightdb/risinglight/blob/8464ae49879b2745f31674dcda43caefd42c86fd/src/main.rs#L131-L152

alissa-tung:
close #485

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -195,12 +195,11 @@ impl sqllogictest::AsyncDB for DatabaseWrapper {
 async fn run_sqllogictest(db: Database, path: &str, output_format: Option<String>) -> Result<()> {
     let mut tester = sqllogictest::Runner::new(DatabaseWrapper { db, output_format });
     let path = path.to_string();
-    let sqllogictest_handler = tokio::task::spawn_blocking(move || {
+    tester
+        .run_file_async(path)
+        .await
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -170,52 +171,36 @@ async fn run_sql(db: Database, path: &str, output_format: Option<String>) -> Res
...
-        while let Some(sql) = trx.recv().await {
-            dtx.send(db.run(&sql).await).await.unwrap();
-        }
-    });
-
+    let mut tester = sqllogictest::Runner::new(DatabaseWrapper { db, output_format });
     let path = path.to_string();
     let sqllogictest_handler = tokio::task::spawn_blocking(move || {
         // `ParseError` isn't Send, so we cannot directly use it as anyhow Error.
         tester.run_file(path).map_err(|err| anyhow!("{:?}", err))?;
```
skyzh, line 200: We can use `run_sync`, and we won't need to spawn tokio future.

alissa-tung: thx

skyzh (approved):
Thanks!

PR #511, opened 2022-02-21
executor: support delete full table, in terms of `delete {} where true`

alissa-tung:
Signed-off-by: alissa-tung <alissa-tung@outlook.com>


skyzh (approved):
Thanks!

Issue #99, opened 2021-11-02
catalog: internal table support


skyzh:
It would be very helpful if we could know internal states of our storage engine using SQL queries. e.g.

```
SELECT rowset_id, size FROM internal.storage.rowsets WHERE table_id = 1;
```

skyzh:
... maybe our first internal table could be `contributor list`?

xxchan:
I'm wondering how to maintain internal tables.. 

I'm considering a `TableType` for internal tables. For example, `contributor list` can be generated in scripts and passed by `env!`, so it is more like a memory table (or just use `InMemoryTable` directly). But the problem is that each storage engine is bound to its `TableType`, so I cannot use an `InMemoryTable` inside `secondary` engine. If we instead implement internal tables differently for each engine, it is not elegant and maybe is inproper (e.g., `contributor list` doesn't need to be stored on disk).

Another idea is to create a separate engine for internal tables, but it seems that we can only use one engine currently?

skyzh:
Internal tables should be a separate engine. e.g., `InternalTableExecutor`, with different endpoints of our system (e.g., compactor, query engine, etc.) providing `get_statistics(&self) -> DataChunk` API to fetch the content of tables.

PR #512, opened 2022-02-21
feat: internal table support (contributors)

xxchan:
close https://github.com/risinglightdb/risinglight/issues/99

```sql
> \dt
+---+----------+---+------------+---+--------------+
| 0 | postgres | 1 | pg_catalog | 0 | contributors |
+---+----------+---+------------+---+--------------+
> select * from pg_catalog.contributors
+--------------+
| name         |
+--------------+
| Alex Chi     |
| Alissa Tung  |
| ArkBriar     |
| Chojan Shang |
| Frank Ma     |
| Lu Di        |
| Mingj        |
| MingjiHan    |
| Runji Wang   |
| Tab Zhang    |
| TennyZhuang  |
| Ting Sun     |
| Yingjun Wu   |
| Yuhui Liu    |
| Yujia Qiao   |
| Zehao Wei    |
| kutori       |
| likg227      |
| nanderstabel |
| stonepage    |
| wzl          |
| xxchan       |
| xyz          |
| 庄天翼       |
|              |
+--------------+
```

```sql
> select count(*) from pg_catalog.contributors
+-------+
| count |
+-------+
| 25    |
+-------+
```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -4,8 +4,30 @@ use std::env;
 use std::fmt::Write as _;
 use std::io::Write;
 use std::path::PathBuf;
+use std::process::Command;
+
+pub fn add_env_contributors() {
+    let output = Command::new("git")
+        .arg("shortlog")
+        .arg("HEAD")
+        .arg("-s")
+        .output()
+        .unwrap();
+    let authors: String = String::from_utf8(output.stdout)
+        .unwrap()
+        .lines()
+        .map(|s| {
+            let mut s = s[7..].to_owned();
+            s.push(',');
+            s
+        })
+        .collect();
+    println!("cargo:rerun-if-changed=.git/HEAD");
+    println!("cargo:rustc-env=RISINGLIGHT_CONTRIBUTORS={}", authors);
+}
 
 fn main() {
+    add_env_contributors();
     // Scan test scripts and generate test cases.
     println!("cargo:rerun-if-changed=tests/sql");
     const PATTERN: &str = "tests/sql/**/[!_]*.slt"; // ignore files start with '_'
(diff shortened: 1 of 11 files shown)


xxchan:
Btw, the generate contributors list contains duplicate names 😅

skyzh (commented):
LGTM! But I prefer not to add too much to our `build.rs`... Now every time we checkout to a new branch or add a new commit, we will trigger a full rebuild...

I think it's better to store contributor list to a JSON, and update it on every release.

xxchan:
> Now every time we checkout to a new branch or add a new commit, we will trigger a full rebuild...

Yes I also have this concern😓😓😓

> I think it's better to store contributor list to a JSON, and update it on every release.

Maybe generated by a pre-release script? (and now we can use this api https://api.github.com/repos/risinglightdb/risinglight/contributors)
Is there something like this? (I don't know much about how to release)

skyzh:
Maybe we can take a look at how https://github.com/datafuselabs/databend update their contributor table?

xxchan:
I actually looked. They use build.rs 😸

build.rs
```diff
@@ -4,8 +4,30 @@ use std::env;
...
+        .arg("HEAD")
+        .arg("-s")
+        .output()
+        .unwrap();
+    let authors: String = String::from_utf8(output.stdout)
+        .unwrap()
+        .lines()
+        .map(|s| {
+            let mut s = s[7..].to_owned();
+            s.push(',');
```
xxchan, line 21: buggy, there's a trailing ',' at the end 🤪🤪🤪

skyzh (approved):

PR #515, opened 2022-02-22
minor fix: update `sqllogictest` pkg remote url

alissa-tung:
Signed-off-by: alissa-tung <alissa-tung@outlook.com>


TennyZhuang (approved):
Thanks!

PR #528, opened 2022-02-24
fix(storage): fix bug of RleBlockIterator.skip()

### LINKED ISSUE #525: storage: wrong result when RLE encoding is enabled
When https://github.com/risinglightdb/risinglight/pull/507 gets merged, and `is_rle = true` for default cli storage options, TPC-H will produce wrong results in some cases. This will only happen when filter scan is triggered.

The issue lies in `skip` implementation of block.

```rust
    fn skip(&mut self, cnt: usize) {
        let mut cnt = cnt;
        while cnt > 0 {
            let rle_count = self.get_cur_rle_count();
            let cur_left = rle_count as usize - self.cur_scanned_count;
            if cur_left > cnt {
                self.cur_scanned_count += cnt;
                self.row_scanned_count += cnt;
                return;
            } else {
                cnt -= cur_left;
                self.row_scanned_count += cur_left;
                self.cur_scanned_count = 0;
                self.cur_row += 1;
                self.block_iter.skip(1);
                if self.cur_row >= self.rle_row_count {
                    break;
                }
            }
        }
    }
```

When we move to the next block on `self.block_iter.skip(1);`, we cannot simply skip. Consider the following cases: we have a RLE block with [1 (repeat 60), 2 (repeat 100), 3 (repeat 50)]:

```rust
iter.next_batch(50);
iter.skip(100);
iter.next_batch(50);
```

With the current logic, the second call to `next_batch` would produce 1.

cc @ludics would you please help fix the bug, and add unit tests for this case (skip function)?

ludics:
close #525 

Signed-off-by: ludics <leonludics@gmail.com>

### DIFF
--- a/src/storage/secondary/block/rle_block_iterator.rs
+++ b/src/storage/secondary/block/rle_block_iterator.rs
@@ -80,6 +80,14 @@ where
         let mut rle_buffer = &self.rle_block[self.cur_row * std::mem::size_of::<u16>()..];
         rle_buffer.get_u16_le()
     }
+
+    fn get_cur_element(&mut self) -> Option<Option<<A::Item as ToOwned>::Owned>> {
+        let mut array_builder = A::Builder::new();
+        if self.block_iter.next_batch(Some(1), &mut array_builder) == 0 {
+            return None;
+        }
+        Some(array_builder.finish().get(0).map(|x| x.to_owned()))
+    }
 }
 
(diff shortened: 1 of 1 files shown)


skyzh (approved):
LGTM, good work!

skyzh:
get_cur_element -> get_next_element?

PR #529, opened 2022-02-24
chore(storage): larger cache and smaller block size

skyzh:

Previously, the block cache is too small. At the same time, the block size is a somehow too large that filter scan barely takes effect in TPC-H queries.


zzl200012 (approved):
LGTM

PR #530, opened 2022-02-24
chore(storage): use tokio OwnedMutexGuard

skyzh:

It turned out that tokio already has an `OwnedMutexGuard`... Before realizing tokio already provides such functionality, I wasted ~1hr implementing a self-referential struct and wrapping MutexGuard with a Arc by myself 🤣


wangrunji0408 (approved):
LGTM

Issue #526, opened 2022-02-23
planner: support group by expression


skyzh:
```
select v2 + 1, sum(v1) from t1 group by v2 + 1;
```

... which currently yields:

```
thread 'tokio-runtime-worker' panicked at 'Plus(Column #ColumnRefId { database_id: 0, schema_id: 0, table_id: 8, column_id: 1 }, Int32(1) (const)) cannot be a group key', src/optimizer/logical_plan_rewriter/input_ref_resolver.rs:111:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'failed to join query thread: JoinError::Panic(...)', src/main.rs:106:23
```

Fedomn:
I'd like to try this

skyzh:
This would be hard... I haven't thought of how to achieve this. I think the first problem you will meet is in `InputRefResolver`.

```
select v2 + 1, sum(v1) from t1 group by v2 + 1;
```

In `InputRefResolver`, we will need to bind a `InputRef` for HashAgg. However, in the case of this query, the binding logic should be non-trivial.

I think supporting group-by expression might need a refactor across the binder, planner and optimizer. If you have any idea, I'd suggest first posting it in Slack group / Discussion or this issue.

Thanks @Fedomn !

jiangzhe:
IMO, it's a little tricky to achieve correct name resolution of `GROUP BY` clause(as well as `HAVING`, not mentioned here).
Basically, the resolution logic could follow `WHERE` or `SELECT` clause, which use table list specified in `FROM`.
Then there are two ways to implement the specific logic of `GROUP BY`.
1. Try match entire expression with `SELECT`ed expression list, prior to normal resolution.
2. Instead of changing resolution logic, add post-validation checking whether the non-aggregate expression is in `SELECT`ed list.
One more thing to note: the actual implementations of different databases diverge on how to resolve group items, especially when groups and aggregate functions are mixed in the same expression.
Take MySQL and Oracle as examples.

```sql
CREATE TABLE t1 (c0 int, c1 int);
SELECT c0, c0 + count(*) FROM t1 GROUP BY c0; -- succeeds in both
SELECT c0, c0+1 +count(*) from t1 GROUP BY c0; -- succeeds in both
SELECT c0+1, c0+1 + count(*) FROM t1 GROUP BY c0+1; -- fails in MySQL but succeeds in Oracle
[...]

PR #535, opened 2022-02-27
feat(executor): support group by expression

Fedomn:
try to close https://github.com/risinglightdb/risinglight/issues/526

### DIFF
--- a/src/binder/expression/column_ref.rs
+++ b/src/binder/expression/column_ref.rs
@@ -9,7 +9,6 @@ use crate::catalog::ColumnDesc;
 /// A bound column reference expression.
 #[derive(PartialEq, Clone, Serialize)]
 pub struct BoundColumnRef {
-    pub table_name: String,
     pub column_ref_id: ColumnRefId,
     pub is_primary_key: bool,
     pub desc: ColumnDesc,
@@ -35,7 +34,6 @@ impl Binder {
                     col.desc().clone(),
                 );
                 let expr = BoundExpr::ColumnRef(BoundColumnRef {
-                    table_name: table.name().clone(),
                     column_ref_id,
                     is_primary_key: col.is_primary(),
                     desc: col.desc().clone(),
@@ -70,7 +68,6 @@ impl Binder {
             let column_ref_id = ColumnRefId::from_table(table_ref_id, col.id());
             self.record_regular_table_column(name, column_name, col.id(), col.desc().clone());
             Ok(BoundExpr::ColumnRef(BoundColumnRef {
-                table_name: name.clone(),
                 column_ref_id,
                 is_primary_key: col.is_primary(),
                 desc: col.desc().clone(),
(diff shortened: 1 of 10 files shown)


skyzh (approved):
LGTM, thanks a lot for this big feature!

st1page (approved):
LGTM, thanks for your contribution!
BTW, I think we should refactor the `Bound::Expr` some day, because it seems to cause our some trouble...

skyzh:
Yes, we are implementing too many things on `Bound::Expr`... Maybe we can implement some visitors for expr :)

xxchan:
The header is incorrect
```sql
> SELECT c0+1, c0+1+count(*) FROM t1 GROUP BY c0+1; 
+----+----------+
| c0 | ?column? |
+----+----------+
| 4  | 5        |
| 2  | 3        |
+----+----------+
in 0.004s
```


This reports error
```sql 
> SELECT c0+2 FROM t1 GROUP BY c0+1; 
logical plan error: (Column #ColumnRefId { database_id: 0, schema_id: 0, table_id: 2, column_id: 0 } + 2) must appear in the GROUP BY clause or be used in an aggregate function
```
But this panics
```sql
> SELECT c0+2+count(*) FROM t1 GROUP BY c0+1; 
thread 'tokio-runtime-worker' panicked at 'Column #ColumnRefId { database_id: 0, schema_id: 0, table_id: 2, column_id: 0 } should not be evaluated in `eval_array`', src/executor/evaluator.rs:55:18

**Comment by xxchan** [comment]
```sql
> select c0*1 from t1 group by c0*1
+----+
| c0 |
+----+
| 3  |
| 1  |
+----+
in 0.004s
> select c0*0 from t1 group by c0*0
thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 0 but the index is 0', src/optimizer/plan_nodes/logical_projection.rs:66:25
```

Fedomn:
Thanks for your input!

> But this panics
> 
> ```sql
> > SELECT c0+2+count(*) FROM t1 GROUP BY c0+1; 
> thread 'tokio-runtime-worker' panicked at 'Column #ColumnRefId { database_id: 0, schema_id: 0, table_id: 2, column_id: 0 } should not be evaluated in `eval_array`', src/executor/evaluator.rs:55:18
> ```

... it seems should report logical plan error that same behavior as postgres



> ```sql
> > select c0*1 from t1 group by c0*1
> +----+
> | c0 |
> +----+
> | 3  |
> | 1  |
> +----+
> in 0.004s
> > select c0*0 from t1 group by c0*0
> thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 0 but the index is 0', src/optimizer/plan_nodes/logical_projection.rs:66:25
> ```

... it seems to be a bug. 
Because after `ArithExprSimplificationRule` rewrite_expr, `c0*0` will rewrite to `Constant(Int32(0))` and the LogicalProjection's child_schema will be empty, however `project_expressions` already replaced to `InputRef #0`, so when get projection schema caused panic.
currently, I'm not sure how to resolve this, maybe just return `"?column?"` when child schema is empty?

PR #536, opened 2022-02-28
feat(executor): use ValueExecutor for simple queries

### LINKED ISSUE #508: executor: use `ValuesExecutor` for simple queries
```
> explain select 1
PhysicalProjection:
    1
  Dummy:
```

Currently, we are using Projection executor for `select 1`, which requires `DummyExecutor` to return one tuple. This is causing problems for many cases. For example,

```
> explain select count(*) from t1 where 0=1
PhysicalProjection:
    InputRef #0
  PhysicalSimpleAgg:
      count(InputRef #0) -> INT
    Dummy:
> select count(*) from t1 where 0=1
+-------+
| count |
+-------+
| 1     |
+-------+
```

`count(*) == 1`, while the filter condition is always false...

Therefore, the first step is to use `ValuesExecutor` for simple queries. After that, we may look into the issue of `count(*)`.

WindowsXp-Beta:
fix #508


src/logical_planner/select.rs
```diff
@@ -44,6 +44,16 @@ impl LogicalPlaner {
                 }
             }
             plan = self.plan_table_ref(table_ref, with_row_handler, is_sorted)?;
+        } else {
+            plan = Arc::new(LogicalValues::new(
+                stmt.select_list
+                    .iter()
+                    .map(|expr| expr.return_type().unwrap())
+                    .collect_vec(),
+                vec![],
```
skyzh, line 53: Considering the column desc is not used any more, what about removing it from `ValuesExecutor`?

skyzh (commented):
Rest LGTM

src/logical_planner/select.rs
WindowsXp-Beta: ok. I'll remove it when I get to school.

skyzh (approved):
LGTM! You may get this merged once the commits are resolved.

WindowsXp-Beta:
It needs those with write access to merge PR.

xxchan:
You need to sign DCO 🐱

WindowsXp-Beta:
thx

xxchan:
@WindowsXp-Beta  You can add your git email to your GitHub account, so that it can be recognized

WindowsXp-Beta:
You mean replace the gmail on my profile with singularity's outlook email?

xxchan:
No, you can add multiple emails. The primary email doesn't need to be changed

WindowsXp-Beta:
ok, added

PR #540, opened 2022-03-01
chore: replace match option with map+transpose

xxchan:
Does this looks good……? 😅


TennyZhuang (approved):
LGTM!

wangrunji0408:
I felt it's better after I learned the `transpose` API. 😅

PR #543, opened 2022-03-02
refactor: introduce visitor for expr

xxchan:
- Introduce `ExprVisitor`, `ExprRewriter` (aka `ExprVisitorMut`?) to reduce duplicated traversing code. 
> Yes, we are implementing too many things on `BoundExpr`... Maybe we can implement some visitors for expr :)
>
> _Originally posted by @skyzh in https://github.com/risinglightdb/risinglight/issues/535#issuecomment-1055151048_
- Optimizer: Add a rule to merge nested projection & Implement `prune_col` for filter (which used the visitor)

### DIFF
--- a/src/optimizer/plan_nodes/logical_filter.rs
+++ b/src/optimizer/plan_nodes/logical_filter.rs
@@ -1,8 +1,6 @@
 // Copyright 2022 RisingLight Project Authors. Licensed under Apache-2.0.
 
-
 use std::fmt;
-
 
 use serde::Serialize;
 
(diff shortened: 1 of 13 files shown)


xxchan:
CI failed once when downloading nextest 😠
https://github.com/risinglightdb/risinglight/runs/5397711071?check_suite_focus=true

skyzh (approved):
Thanks!

src/optimizer/plan_nodes/logical_filter.rs
```diff
@@ -52,10 +52,116 @@ impl PlanNode for LogicalFilter {
...
+        impl ExprRewriter for ShiftLeft {
+            fn rewrite_input_ref(&self, expr: &mut BoundExpr) {
+                match expr {
+                    BoundExpr::InputRef(ref mut input_ref) => {
+                        input_ref.index -= self.0;
+                    }
+                    _ => unreachable!(),
+                }
+            }
+        }
```
xxchan, line 76: Shift is not enough: input refs `[1,3]` should become `[0,1]`. But since we don't have sub-queries like `select v1, v2 from (select v1,v3 from t)`, this is hard to trigger. 🤪

PR #544, opened 2022-03-02
WIP storage: checksum BlockType in block header

### LINKED ISSUE #247: storage: checksum `BlockType` in block header
Currently, we store the `BlockType` in header of each block. Therefore, this part is not checksumed. We should get this part checksumed.

hawkingrei:
close #247

Signed-off-by: Weizhen Wang <wangweizhen@pingcap.com>

### DIFF
--- a/src/storage/secondary/block/block_index_builder.rs
+++ b/src/storage/secondary/block/block_index_builder.rs
@@ -1,5 +1,6 @@
 // Copyright 2022 RisingLight Project Authors. Licensed under Apache-2.0.
 
+use bytes::BufMut;
 use risinglight_proto::rowset::block_index::BlockType;
 use risinglight_proto::rowset::{BlockIndex, BlockStatistics};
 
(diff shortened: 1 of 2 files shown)


xxchan (commented):
I think calculate checksum field by field is not very elegant because we may need to add more fields later. Checksum a whole chunk of bytes may be better

> I think we may need to change the encoding of block headers (e.g., put block type and checksum at the end?) in order to support this.
>
> _Originally posted by @skyzh in https://github.com/risinglightdb/risinglight/issues/247#issuecomment-1024847607_

hawkingrei:
OK，let me research it.
> I think calculate checksum field by field is not very elegant because we may need to add more fields later. Checksum a whole chunk of bytes may be better
> 
> 
> 
> > I think we may need to change the encoding of block headers (e.g., put block type and checksum at the end?) in order to support this.
> 
> >
> 
> > _Originally posted by @skyzh in https://github.com/risinglightdb/risinglight/issues/247#issuecomment-1024847607_

PR #548, opened 2022-03-04
feat: map filter inputref

### LINKED ISSUE #547: optimizer: mapping `InputRef` when pruning columns
When pruning columns, we need to map `InputRef`. 

e.g., pruning
```text
Filter(cond: input_ref(1)<5)
  TableScan(v1, v2, v3)
```
with required columns `[2]` will result in
```text
Filter(cond: input_ref(0)<5)  # 1->0
  TableScan(v2, v3)
```

Currently we shifted them (subtract each one with an offset), which turns the smallest one into `0`: https://github.com/risinglightdb/risinglight/pull/543/files#diff-eb434564a52231d1f4486e36fdab58dc83a6709af14dc87554def86bdc4ecfe9R66-R76 

But shift is not enough: inconsecutive input refs `[1,3]` should become `[0,1]`. But since we don't have sub-queries like `select v1, v2 from (select v1,v3 from t)`, this is hard to trigger.

_Originally posted by @xxchan in https://github.com/risinglightdb/risinglight/pull/543#r818370427_

XieJiann:
Signed-off-by: xiejiann <jianxie0@gmail.com>
close #547 

Map the index when pruning cols in the filter

### DIFF
--- a/src/optimizer/plan_nodes/logical_filter.rs
+++ b/src/optimizer/plan_nodes/logical_filter.rs
@@ -84,7 +84,7 @@ impl PlanNode for LogicalFilter {
         }
 
         let mut expr = self.expr.clone();
-        Mapper(idx_table).rewrite_expr(&mut expr);
+        Mapper(idx_table.clone()).rewrite_expr(&mut expr);
 
         let new_filter = Self {
             expr,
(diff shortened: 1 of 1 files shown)


xxchan (approved):
Nice work! LGTM but let @st1page also take a look

Issue #537, opened 2022-02-28
storage: use object to pin and unpin version


skyzh:
https://github.com/risinglightdb/risinglight/blob/7b5bc7a1a258afd64a483990ca95f4c799482791/src/storage/secondary/version_manager.rs#L258-L263

Currently, users will need to manually call `pin` and `unpin`. If there are futures cancelled or somehow, it is possible that we fail to unpin by ourselves.

We should implement a struct called `Version`, and implement `Drop` for it. `pin` will return a `Arc<Version>` instead of some strange things. When a `Version` gets dropped (which means that all reference of `Arc<Version>` is now dropped), we can call `unpin` in `Drop`.

skyzh:
```rust
pub struct Version(u64, Arc<Snapshot>);

impl Drop for Version {
 // ...
}

pub fn pin(&self) -> Arc<Version> { /* .. */ }
```

Yisaer:
I'm not familiar with rust, May I try this issue? :)

skyzh:
Sure, thanks for that!

PR #549, opened 2022-03-04
storage: use Arc<Version> during pin and unpin

Yisaer:
Signed-off-by: yisaer <disxiaofei@163.com>

close https://github.com/risinglightdb/risinglight/issues/537

Currently, we use `pin` and `unpin` manually in code. In this request, we use `Arc<Version>` to help us automatically unpin instead of manual code.

### DIFF
--- a/src/storage/secondary/transaction.rs
+++ b/src/storage/secondary/transaction.rs
@@ -45,9 +45,6 @@ pub struct SecondaryTransaction {
     /// Snapshot content
     snapshot: Arc<Snapshot>,
 
-    /// Epoch of the snapshot
-    epoch: u64,
-
     /// The rowsets produced in the txn.
     to_be_committed_rowsets: Vec<DiskRowset>,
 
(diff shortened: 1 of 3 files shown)


src/storage/secondary/transaction.rs
```diff
@@ -70,15 +67,15 @@ impl SecondaryTransaction {
         update: bool,
     ) -> StorageResult<Self> {
         // pin a snapshot at version manager
-        let (epoch, snapshot) = table.version.pin();
+        let version = table.version.pin();
```
skyzh, line 70: Should store version inside txn struct, otherwise it will get immediately unpinned once txn is created.

skyzh (commented):
Rest LGTM, thanks!

I've another idea: we can simply using Arc to help us do reference counting, instead of maintaining the ref_cnt by ourselves. For each epoch, we associate a single Version object, and wrap it with an Arc. When a user requests pin, we clone the Arc for them. When the last Arc gets dropped, the Version object gets dropped, and we can safely remove the version.

skyzh:
solved by https://github.com/risinglightdb/risinglight/pull/685

Issue #551, opened 2022-03-07
storage: use `get_or_insert_with` for block cache


skyzh:
Currently, we first `get` and then fill the cache. Therefore, if there are multiple requests to a single block, while the block is not in cache, we will issue multiple requests to read from disk.

https://github.com/risinglightdb/risinglight/blob/53ccdf96f36267c4703386577eeeff1da18bdf13/src/storage/secondary/column.rs#L154-L187

By using `get_or_insert_with` interface, we can use `moka` cache to help solve this problem.

BaymaxHWY:
Hi, may I try this issue? just use the `get_or_insert_with` method to prevent multiple reads from disk?(one read from disk, other reads from cache)
Are there any test cases that can trigger this problem?

skyzh:
Thanks! No need to add unit tests, as we currently can only run one query at a time.

PR #556, opened 2022-03-08
chore(storage): use get_or_try_insert_with for block cache

BaymaxHWY:
fix: #551

### DIFF
--- a/src/storage/secondary/column.rs
+++ b/src/storage/secondary/column.rs
@@ -182,8 +182,7 @@ impl Column {
                 do_verify_checksum = true;
                 block
             })
-            .await
-            .unwrap();
+            .await?;
 
         if block.len() < BLOCK_HEADER_SIZE {
             return Err(TracedStorageError::decode(
--- a/src/storage/secondary/column.rs
+++ b/src/storage/secondary/column.rs
@@ -149,42 +149,41 @@ impl Column {
         let key = self.base_block_key.clone().block(block_id);
 
         let mut block_header = BlockHeader::default();
-        let do_verify_checksum;
+        let mut do_verify_checksum = false;
 
-        let block = if let Some(block) = self.block_cache.get(&key) {
-            // No need to verify checksum when read from cache
-            do_verify_checksum = false;
-            block
-        } else {
-            // block has not been in cache, so we fetch it from disk
-            // TODO(chi): support multiple I/O backend
+        // support multiple I/O backend
+        let block = self
+            .block_cache
+            .get_or_try_insert_with(key, async {
+                // block has not been in cache, so we fetch it from disk
+                let file = self.file.clone();
+                let info = self.index.index(block_id).clone();
+                let block = tokio::task::spawn_blocking(move || {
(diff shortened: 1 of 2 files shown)


src/storage/secondary/column.rs
```diff
@@ -149,42 +149,41 @@ impl Column {
...
             .await
-            .unwrap()?;
-
-            // TODO(chi): we should invalidate cache item after a RowSet has been compacted.
-            self.block_cache.insert(key, block.clone()).await;
-
-            do_verify_checksum = true;
-            block
-        };
+            .unwrap();
```
skyzh, line 186: Is there anyway to return this error? e.g. add a `Arc<StroageError>` variant for `StorageError`?

skyzh (approved):
Generally LGTM, good work!

src/storage/secondary/column.rs
BaymaxHWY: i'm sorry, i not very clear about "add a `Arc<StroageError>` variant for `StorageError`", i tried to use `?` replace `unwrap`, but this need impl `From` trait(I'm not particularly familiar with this part)

It's all I can think of to do
```rust
let block = match block {
    Ok(block) => block,
    Err(_) => return Err(TracedStorageError::not_found("block", block_id)),
};
```
Maybe I need some examples...[rust rookie :)]

skyzh: https://github.com/risinglightdb/risinglight/blob/733be2953af4b434ccd0c59f40aa71f638d4887d/src/storage/error.rs#L9-L33

Here we can add a new nested enum variant:

```rust
    #[error("{0}")]
    #[backtrace]
    #[from]
    Nested(Arc< StorageError>),
}
```

So that we can directly use `?` on such error.

BaymaxHWY: and the `get_or_try_insert_with` will be panics when the `init` future has been panicked, Whether a file system error causes panic?

BaymaxHWY: ok, i will to fix it after i finish this PR, hhh!

BaymaxHWY: Hey, I went to look at the code and I think it should return error instead of panic, the file system error will pass to `block` of `spawn_blocking` and pass to `block` of `get_or_try_insert_with`.
Of course, if a file system panic occurs, the task calling this method will panic, other task will continue work.

skyzh: Agree. You may send a new PR, thanks!

BaymaxHWY: > Hey, I went to look at the code and I think it should return error instead of panic, the file system error will pass to `block` of `spawn_blocking` and pass to `block` of `get_or_try_insert_with`. Of course, if a file system panic occurs, the task calling this method will panic, other task will continue work.

Sorry, I may have a problem with my expression (forgive my broken English), I mean it can do that now. Don't to change any code, just remove the `TODO` comment :)

skyzh: Sorry, I misunderstood you. A new PR to remove `TODO` is welcomed!

PR #558, opened 2022-03-10
fix(executor): Char and Varchar's length limit not respected

### LINKED ISSUE #455: Char and Varchar's length limit not respected
Secondary engine 
```
> create table c (x char(3) not null);
created
> insert into c values ('111')
1 rows inserted
> insert into c values ('1111')
thread 'main' panicked at 'item length 4 > char width 3', src/storage/secondary/block/char_block_builder.rs:39:13

> create table cc (x varchar(3) not null)
created
> insert into cc values ('1111')
1 rows inserted
```

https://github.com/risinglightdb/risinglight/blob/c221b085ceb4a61fe1cf1e6c9d6caa36b7215a2e/src/storage/secondary/column/column_builder.rs#L44

Because varchar has a length limit...

---

memory engine has no check at all
```
> create table cc (x char(3) not null)
created
> insert into cc values ('1111')
1 rows inserted
```

xiaoyong-z:

close #455 

1. add char and varchar length check in  ValuesExecutor

### DIFF
--- a/src/executor/values.rs
+++ b/src/executor/values.rs
@@ -37,7 +37,13 @@ impl ValuesExecutor {
                         _ => None,
                     };
                     if let Some(width) = size {
-                        let item_length = value.get(0).len() as u64;
+                        let item_length = if let DataValue::String(x) = &value.get(0) {
+                            x.len() as u64
+                        } else if let DataValue::Null = &value.get(0) {
+                            0
+                        } else {
+                            unreachable!()
+                        };
                         if item_length > width {
                             return Err(ExecutorError::ExceedLengthLimit {
                                 length: item_length,
--- a/src/executor/values.rs
+++ b/src/executor/values.rs
@@ -31,20 +31,17 @@ impl ValuesExecutor {
             for row in chunk {
                 for (expr, column_type, builder) in izip!(row, column_types, &mut builders) {
                     let value = expr.eval(&dummy)?;
-                    if let Type::Varchar(Some(size)) = column_type.kind {
-                        let item_length = value.get(0).to_string().len() as u64;
-                        if item_length > size {
+                    let size = match column_type.kind {
+                        Type::Varchar(size) => size,
+                        Type::Char(size) => size,
+                        _ => None,
+                    };
+                    if let Some(width) = size {
+                        let item_length = value.get(0).len() as u64;
+                        if item_length > width {
                             return Err(ExecutorError::ExceedLengthLimit {
                                 length: item_length,
-                                width: size,
-                            });
-                        }
-                    } else if let Type::Char(Some(size)) = column_type.kind {
-                        let item_length = value.get(0).to_string().len() as u64;
-                        if item_length > size {
-                            return Err(ExecutorError::ExceedLengthLimit {
-                                length: item_length,
-                                width: size,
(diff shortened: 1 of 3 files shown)


src/executor/values.rs
```diff
@@ -17,17 +19,35 @@ impl ValuesExecutor {
...
                 .map(|ty| ArrayBuilderImpl::with_capacity(chunk.len(), ty))
                 .collect_vec();
             // Push value into the builder.
             let dummy = DataChunk::single(0);
             for row in chunk {
-                for (expr, builder) in row.iter().zip_eq(&mut builders) {
+                for (expr, column_type, builder) in izip!(row, column_types, &mut builders) {
                     let value = expr.eval(&dummy)?;
+                    if let Type::Varchar(Some(size)) = column_type.kind {
+                        let item_length = value.get(0).to_string().len() as u64;
```
skyzh, line 35:
```suggestion
                        let item_length = value.get(0).len() as u64;
```

src/executor/values.rs
```diff
@@ -17,17 +19,35 @@ impl ValuesExecutor {
...
                 .iter()
                 .map(|ty| ArrayBuilderImpl::with_capacity(chunk.len(), ty))
                 .collect_vec();
             // Push value into the builder.
             let dummy = DataChunk::single(0);
             for row in chunk {
-                for (expr, builder) in row.iter().zip_eq(&mut builders) {
+                for (expr, column_type, builder) in izip!(row, column_types, &mut builders) {
                     let value = expr.eval(&dummy)?;
+                    if let Type::Varchar(Some(size)) = column_type.kind {
```
skyzh, line 34: We can firstly store `Option<usize>` in a variable, and then check it. Therefore, we can write the `if item_length > size` part only once.

src/executor/values.rs
```diff
@@ -17,17 +19,32 @@ impl ValuesExecutor {
...
-                for (expr, builder) in row.iter().zip_eq(&mut builders) {
+                for (expr, column_type, builder) in izip!(row, column_types, &mut builders) {
                     let value = expr.eval(&dummy)?;
+                    let size = match column_type.kind {
+                        Type::Varchar(size) => size,
+                        Type::Char(size) => size,
+                        _ => None,
+                    };
+                    if let Some(width) = size {
+                        let item_length = value.get(0).len() as u64;
```
skyzh, line 40: If `width` is `Some`, we are sure that the `value` will be a `String`. Therefore, we can:

let item_length = if let DataValue::String(x) = value.get(0) { x.len() as u64 } else if let DataValue::Null = value.get(0) { 0 } else { unreachable!() };

src/types/mod.rs
```diff
@@ -296,6 +296,23 @@ impl DataValue {
             }
         }))
     }
+
+    /// Return the len of value, the length for [Decimal| Date| Interval] is based on the number of
+    /// i32 element variable it contains.
+    pub fn len(&self) -> usize {
```
skyzh, line 302: `len` of a value doesn't seem apparent to me... Only `String` has length.

skyzh (commented):
Thanks for your contribution!

skyzh (approved):
Thanks!

Issue #435, opened 2022-02-09
chore: separate output header from `DataChunk`


pleiadesian:
According to [this comment](https://github.com/risinglightdb/risinglight/pull/429#pullrequestreview-876927077), we should not bind `header` with `DataChunk`. As `header` is only used as output column names, we can add a `ResultChunk`, which wraps the output header and output `DataChunk`.

RinChanNOWWW:
I can have a try.

PR #560, opened 2022-03-12
chore: remove header in `DataChunk`.

RinChanNOWWW:
Signed-off-by: RinChanNOWWW <hzy427@gmail.com>

close https://github.com/risinglightdb/risinglight/issues/435

Remove the member `header` in the struct `DataChunk`. And it seems that there is no need to add a wrapper struct `ResultChunk`. Adding `ResultChunk` will complicate the executors.


skyzh (approved):
LGTM, thanks!

PR #564, opened 2022-03-15
do not detect tests in build.rs

### LINKED ISSUE #472: tests: small change in `tests` folder will cause a full rebuild of binary
Cause by https://github.com/risinglightdb/risinglight/pull/376

Is there any way that `build.rs` only detects only file name instead of full file content?

xxchan:
Signed-off-by: xxchan <xxchan22f@gmail.com>

in `sqllogictest.rs` instead

closes #472


skyzh (approved):
LGTM, but I'd prefer tests could be running in parallel and test names can be shown in stdout...

Issue #565, opened 2022-03-15
test: parallelize sqllogictest


xxchan:
https://github.com/risinglightdb/risinglight/pull/564

PR #568, opened 2022-03-18
feat(ci): parallism test

skyzh:

close https://github.com/risinglightdb/risinglight/issues/565


xxchan (approved):
😍😍😍

xxchan:
Still not fast enough? 🤪

skyzh:
> Still not fast enough? 🤪

Yes. I think it's better to re-introduce the `build.rs`, but in a separate crate. e.g. `risinglight-sqllogictest` crate. Therefore, we can avoid re-building the whole binary if only test cases get changed.

xxchan:
~Why we didn't come up with that earlier?~ 

Oh I missed the discussion 🤪
> How about moving the integration test into a separate crate?

_Originally posted by @Fedomn in https://github.com/risinglightdb/risinglight/issues/472#issuecomment-1043775252_

skyzh:
Apart from separating a `risinglight-sqllogictest` crate, we also added a new in-memory file backend, which speeds up 2x on each disk unit tests :)

Issue #574, opened 2022-03-21
Tracking: publish to crates.io


skyzh:
Currently, we are using many crates that are installed from git. We need to switch to a published crates.io version before we can publish RisingLight to crates.io.

xxchan:
- sqllogictest: no problem
- sqlparser: our fork made a change: https://github.com/risinglightdb/sqlparser-rs/commit/edead1b
- futures_async_stream: not sure why a commit is used (from #271)

skyzh:
The latest published version of futures-async-stream has different interface from its main branch... so we may wait for the upstream to get things published.

xxchan:
So how to handle `sqlparser`? Do we send PR to upstream or publish fork? cc @wangrunji0408

wangrunji0408:
I'd like to push it upstream. Let me take this.

PR #575, opened 2022-03-21
chore: use published crates

xxchan:
part of #574 
It seems OK?


skyzh (approved):
LGTM! Also sqllogictest is used in tests/sqllogictest, please also help bump it, thanks!

Issue #464, opened 2022-02-14
executor: refactor expression framework to return `Arc<ArrayImpl>`


skyzh:
When grouping by a key, the group expr will be an `InputRef`.

https://github.com/risinglightdb/risinglight/blob/52994366ebcd09047a798d4e452a1989a8568697/src/executor/hash_agg.rs#L31-L33

https://github.com/risinglightdb/risinglight/blob/f75c99bb274dece4215a4ba0cc1b60ff1ab796b8/src/executor/evaluator.rs#L16

And evaluating an `InputRef` will clone the whole Array, which is not efficient.

As this is a common case to solely evaluate an `InputRef`, we can optimize this by changing the expression eval return type to `Arc<ArrayImpl>`.

wangrunji0408:
How about making `ArrayImpl` cheaply clonable (like `bytes::Bytes`)? In other words, put `Arc` inside the array.

skyzh:
> How about making `ArrayImpl` cheaply clonable (like `bytes::Bytes`)? In other words, put `Arc` inside the array.

Good idea. It is designed to be immutable, so it's okay to wrap an Arc inside.

kwannoel:
Would like to take this up!

kwannoel:
Are there any benchmarks you would recommend to demonstrate greater efficiency? Thanks!

kwannoel:
> > How about making `ArrayImpl` cheaply clonable (like `bytes::Bytes`)? In other words, put `Arc` inside the array.
> 
> Good idea. It is designed to be immutable, so it's okay to wrap an Arc inside.

And to clarify is this what you're talking about?
```rs
pub struct ArrayImpl {
    inner: Arc<ArrayImplInner>, // Wrap with Arc, cheap to clone
}

impl ArrayImpl {
    pub fn new(&self, inner: Arc<ArrayImplInner>) -> Self {
        ArrayImpl {inner}
    }
}

#[derive(Debug, Clone, PartialEq)]
pub enum ArrayImplInner {
    Bool(BoolArray),
    // Int16(PrimitiveArray<i16>),
    Int32(I32Array),
    Int64(I64Array),
    // Float32(PrimitiveArray<f32>),
    Float64(F64Array),
    Utf8(Utf8Array),
    Blob(BlobArray),
    Decimal(DecimalArray),
    Date(DateArray),
    Interval(IntervalArray),
}
```

skyzh:
[...]

PR #577, opened 2022-03-21
refactor(executor): refactor ArrayImpl::Bool for cheap cloning

kwannoel:
Part of #464, implements solution for `ArrayImpl::Bool`

Signed-off-by: Noel Kwan [noelkwan1998@gmail.com](mailto:noelkwan1998@gmail.com)

### DIFF
--- a/src/array/mod.rs
+++ b/src/array/mod.rs
@@ -4,6 +4,7 @@ use std::convert::TryFrom;
 use std::fmt::Debug;
 use std::iter::TrustedLen;
 use std::ops::{Bound, RangeBounds};
+use std::sync::Arc;
 
 use rust_decimal::prelude::FromStr;
 use rust_decimal::Decimal;
@@ -161,7 +162,7 @@ pub type IntervalArray = PrimitiveArray<Interval>;
 /// Embeds all types of arrays in `array` module.
 #[derive(Debug, Clone, PartialEq)]
 pub enum ArrayImpl {
-    Bool(BoolArray),
+    Bool(Arc<BoolArray>),
     // Int16(PrimitiveArray<i16>),
     Int32(I32Array),
     Int64(I64Array),
@@ -172,6 +173,12 @@ pub enum ArrayImpl {
     Decimal(DecimalArray),
     Date(DateArray),
     Interval(IntervalArray),
+}
+
+impl ArrayImpl {
+    pub fn new_bool(bool_array: BoolArray) -> Self {
+        Self::Bool(Arc::new(bool_array))
+    }
 }
 
 pub type BoolArrayBuilder = PrimitiveArrayBuilder<bool>;
@@ -234,17 +241,17 @@ macro_rules! impl_from {
             /// Implement `AbcArray -> ArrayImpl`
             impl From<$AbcArray> for ArrayImpl {
                 fn from(array: $AbcArray) -> Self {
-                    Self::$Abc(array)
+                    Self::$Abc(array.into())
                 }
             }
 
             /// Implement `ArrayImpl -> AbcArray`
-            impl TryFrom<ArrayImpl> for $AbcArray {
+            impl TryFrom<ArrayImpl> for Arc<$AbcArray> {
                 type Error = TypeMismatch;
 
                 fn try_from(array: ArrayImpl) -> Result<Self, Self::Error> {
                     match array {
-                        ArrayImpl::$Abc(array) => Ok(array),
+                        ArrayImpl::$Abc(array) => Ok(array.into()),
                         _ => Err(TypeMismatch),
                     }
                 }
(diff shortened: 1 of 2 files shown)


kwannoel:
Will clean-up commits later -- this is just a draft to demo approach for: https://github.com/risinglightdb/risinglight/issues/464

src/array/mod.rs
```diff
@@ -234,7 +243,8 @@ macro_rules! impl_from {
             /// Implement `AbcArray -> ArrayImpl`
             impl From<$AbcArray> for ArrayImpl {
                 fn from(array: $AbcArray) -> Self {
-                    Self::$Abc(array)
+                    // Self::$Abc(array)
```
xxchan, line 246: Can we `Self::$Abc(Arc::new(array))` here? Then no need to add `new_bool`

skyzh: `Self::$Abc(array.into())` might be better. By using `.into()`, we can have both `Arc` and the original representation in such intermediate state.

src/array/mod.rs
```diff
@@ -244,7 +254,8 @@ macro_rules! impl_from {
 
                 fn try_from(array: ArrayImpl) -> Result<Self, Self::Error> {
                     match array {
-                        ArrayImpl::$Abc(array) => Ok(array),
+                        // ArrayImpl::$Abc(array) => Ok(array),
+                        ArrayImpl::$Abc(array) => loop {},
```
skyzh, line 258: I think this function should be implemented on `Arc<Array>` then. `ArrayImpl::Bool` can now only be turned into an `Arc<BoolArray>`.

src/executor/evaluator.rs
```diff
@@ -45,7 +45,7 @@ impl BoundExpr {
             }
             BoundExpr::IsNull(expr) => {
                 let array = expr.expr.eval(chunk)?;
-                Ok(ArrayImpl::Bool(
+                Ok(ArrayImpl::new_bool(
```
skyzh, line 48: ArrayImpl::Bool(<the original BoolArray>.into())

skyzh (commented):
Generally LGTM, and you are on the right direction! Thanks for this PR.

skyzh:
Well, I've tried doing this refactoring on my local branch. It already took me ~20min to doing the refactor (which produces a ~400LoC's diff), and I think I'll take another 20min to complete the whole thing... Some suggestions:

Use `.into()` to transform a AbcArray into Arc<AbcArray>. For example,

```rust
            impl From<$AbcArray> for ArrayImpl {
                fn from(array: $AbcArray) -> Self {
                    Self::$Abc(array.into())
                }
            }
```

Use the `paste` crate to generate boilerplate code:

```rust
            $(
                paste! {
                    /// Create a new array of the corresponding type
                    pub fn [<new_ $abc>](array: $AbcArray) -> Self {
                        ArrayImpl::$Abc(array.into())
                    }
                }
            )*
```

Sometimes the compiler cannot infer type, especially at:

```
ArrayImpl::Bool([v].into_iter().collect() /* cannot determine what to collect! */),
```

We can use `new_bool` to help the compiler to determine type:

```rust
ArrayImpl::new_bool([v].into_iter().collect()),
```

Anyway, nice try for this PR! I think you have already done a good job and on the right direction. You may continue to finish the `BoolArray` part, and if you're interested, benchmark `select bool_column from table`. (However, from my perspective, I think blob array and string array can have more demonstrating benchmark result, as their clone are much more expensive than other primitive arrays.) After the BoolArray change get merged, I can help refactor all other places.

skyzh:
Also please use `panic!("not supported")` or `todo!()` for code that are not implemented... Otherwise CI will spin for too long time :)

image

kwannoel:
> Also please use `panic!("not supported")` or `todo!()` for code that are not implemented... Otherwise CI will spin for too long time :)
> 
> image

Really sorry about that, will do so.

src/executor/evaluator.rs
kwannoel: Leaving this as it uses `collect` to construct array.

src/array/mod.rs
kwannoel: Great suggestion, thanks!

src/array/mod.rs
kwannoel: Done, thanks for suggesting this.

skyzh (approved):
LGTM, thanks! Are you going to add more commits? If not, I'd be happy to merge this.

kwannoel:
> LGTM, thanks! Are you going to add more commits? If not, I'd be happy to merge this.

Nope not adding anymore changes - but feel free to let me know if you'd like anymore help with refactoring the rest of `ArrayImpl`'s members :) I can do so in separate PRs

PR #581, opened 2022-03-23
feat(function): add function definition

MingjiHan99:
Signed-off-by: MingjiHan <mingji_han@outlook.com>

The design refers to Databend. https://github.com/datafuselabs/databend/tree/b22144300e2adaaba81f9cda772f1e00b90176f3/common/functions

TODO: support function binding and execution in our database system.

### DIFF
--- a/src/function/abs.rs
+++ b/src/function/abs.rs
@@ -1,0 +1,91 @@
+use super::*;
+use crate::array::ArrayImpl;
+use crate::function::FunctionError::{InvalidDataTypes, InvalidParameters};
+
+pub struct AbsFunction {}
+
+impl Function for AbsFunction {
+    fn name(&self) -> &str {
+        "abs"
+    }
+
+    fn return_types(
+        &self,
+        input_types: &[PhysicalDataTypeKind],
+    ) -> Result<Vec<PhysicalDataTypeKind>, FunctionError> {
+        if input_types.len() != 1 {
+            return Err(InvalidDataTypes("The column size should be 1".to_string()));
+        }
+        match &input_types[0] {
+            PhysicalDataTypeKind::Int32 => Ok(vec![PhysicalDataTypeKind::Int32]),
+            PhysicalDataTypeKind::Int64 => Ok(vec![PhysicalDataTypeKind::Int64]),
+            PhysicalDataTypeKind::Float64 => Ok(vec![PhysicalDataTypeKind::Float64]),
+            _ => Err(InvalidDataTypes("Data type is not supported".to_string())),
+        }
+    }
+
+    fn execute(&self, input: &DataChunk) -> Result<DataChunk, FunctionError> {
+        if input.column_count() != 1 {
+            return Err(InvalidParameters("The column size should be 1".to_string()));
+        }
+
+        let arr = input.array_at(0);
+        match &arr {
+            ArrayImpl::Int32(i32_arr) => {
+                let mut builder = I32ArrayBuilder::new();
+                for val in i32_arr.iter() {
+                    match val {
+                        Some(val) => {
+                            if *val >= 0 {
+                                builder.push(Some(&*val));
+                            } else {
+                                builder.push(Some(&-*val));
+                            }
+                        }
+                        None => builder.push(None),
+                    }
+                }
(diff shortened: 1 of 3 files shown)


src/function/abs.rs
```diff
@@ -0,0 +1,91 @@
...
+                let res_arr: ArrayImpl = ArrayImpl::from(builder.finish());
+                let vec = vec![res_arr];
+                Ok(vec.into_iter().collect())
+            }
+            ArrayImpl::Int64(i64_arr) => {
+                let mut builder = I64ArrayBuilder::new();
+                for val in i64_arr.iter() {
+                    match val {
+                        Some(val) => {
+                            if *val >= 0 {
```
skyzh, line 57: (*val).abs()?

skyzh (commented):
Rest LGTM. By the way, can we have a more general framework for all functions? If every time we need to add a new function, we need to write such many boilerplate code, this sounds unacceptable.

src/function/abs.rs
MingjiHan99: > Rest LGTM. By the way, can we have a more general framework for all functions? If every time we need to add a new function, we need to write such many boilerplate code, this sounds unacceptable.

The developers could use marco to implement functions, but I think the input types matching is inevtible.
The design actually refers to function in Databend. https://github.com/datafuselabs/databend/tree/b22144300e2adaaba81f9cda772f1e00b90176f3/common/functions

I also checked DuckDB, in which many template tricks are used. I think it is more ticky.

xxchan: I vote for designing a more general framework first, in which we will be more comfortable to implement functions. 

Otherwise the code is doomed to be refactored out, and the hard work is wasted?

MingjiHan99: We can discuss the details about a more general function framework.  
So far I think we can use the current design and continue develop the function binding and executor.  
We will not implement too many functions before our discussion.

src/function/abs.rs
```diff
@@ -0,0 +1,73 @@
...
+use crate::function::FunctionError::{InvalidDataTypes, InvalidParameters};
+
+pub struct AbsFunction {}
+
+impl Function for AbsFunction {
+    fn name(&self) -> &str {
+        "abs"
+    }
+
+    fn return_types(
```
skyzh, line 12: Better to calculate return types in advance when calling new.

src/function/abs.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    fn execute(&self, input: &DataChunk) -> Result<DataChunk, FunctionError> {
+        if input.column_count() != 1 {
+            return Err(InvalidParameters("The column size should be 1".to_string()));
+        }
+
+        let arr = input.array_at(0);
+        match &arr {
+            ArrayImpl::Int32(i32_arr) => {
```
skyzh, line 34: I'd prefer having a framework to vectorize and dispatch by type. Need more consideration.

src/function/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+pub trait Function {
+    // Each function should have an unique name.
+    fn name(&self) -> &str;
+    // A function could support mutiple kinds of data types.
+    // For example, the abosulte value function can get absolute value of integer, float number or
+    // double number.
+    fn return_types(
+        &self,
+        input_types: &[PhysicalDataTypeKind],
+    ) -> Result<Vec<PhysicalDataTypeKind>, FunctionError>;
```
skyzh, line 26: Why there will be multiple returned data types?

skyzh (commented):
Rest LGTM

skyzh:
Also IIRC we already have array.unary_op, which can greatly simplify expression implementation.

src/function/mod.rs
MingjiHan99: My bad:)

src/function/abs.rs
MingjiHan99: Yes, we should consider it carefully. I will not implement more functions before our discussion.

skyzh (approved):
Please resolve above comments before merging 🥰🥰

xxchan:
https://github.com/risinglightdb/risinglight/issues/584 (add a link reference)

PR #586, opened 2022-03-26
feat(tracing): make minitrace  a cli option

D2Lark:
close https://github.com/risinglightdb/risinglight/issues/582

PR #587, opened 2022-03-27
chore(tracing): remove enable_tracing option from db interface

skyzh:

TL;DR, the db interface itself should not expose `enable_tracing`. Tracing is enabled automatically if there's a parent tracing span in current thread.

Also remove fsync from manifest if is in-memory backend. 2x faster for storage e2e tests.

ref https://github.com/risinglightdb/risinglight/pull/586


wangrunji0408 (approved):
LGTM

PR #594, opened 2022-03-30
feat(storage): report statistics after finishing table scan (#474)

### LINKED ISSUE #474: storage: report statistics after finishing table scan
After we have added filter scan functionality, we should know how many blocks have been scanned in a table scan, so as to know whether the filter scan actually takes effect. It should be printed as debug message.

In https://github.com/risinglightdb/risinglight/blob/main/src/storage/secondary/column/concrete_column_iterator.rs, we should:

* record how many times `next_batch` has been called in the struct.
* record how many blocks have been loaded from disk. The following two parts are the logic of fetching data (`get_iterator_for`, which is equivalent to creating new real block iterators.)
https://github.com/risinglightdb/risinglight/blob/6ff143ed6d4555318886284a9c7ad4b0e7d70f0f/src/storage/secondary/column/concrete_column_iterator.rs#L96-L105
https://github.com/risinglightdb/risinglight/blob/6ff143ed6d4555318886284a9c7ad4b0e7d70f0f/src/storage/secondary/column/concrete_column_iterator.rs#L131-L136
* report `next_batch_count` and `fetched_block / total_block` in debug message using `tracing::debug` (which can be done in `impl Drop for ConcreteColumnIterator` or by adding a new function and call it in table scan executor.

adlternative:
Hi, I am a newbie for risinglight and rust :) 

The current patch want to fix #474.

 
Now it simply outputs the next_batch_count in ConcreteColumnIterator drop() function.

```sh
> select sum(L_LINENUMBER) from LINEITEM;
...
2022-03-30T02:51:14.669643Z DEBUG risinglight::storage::secondary::column::concrete_column_iterator: Statistics {
    next_batch_count: 195,
}
2022-03-30T02:51:14.669749Z DEBUG risinglight::storage::secondary::column::concrete_column_iterator: Statistics {
    next_batch_count: 913,
}
2022-03-30T02:51:14.669763Z DEBUG risinglight::storage::secondary::column::concrete_column_iterator: Statistics {
    next_batch_count: 913,
}
2022-03-30T02:51:14.669775Z DEBUG risinglight::storage::secondary::column::concrete_column_iterator: Statistics {
    next_batch_count: 913,
}
```

How can I calculate fetched_block and  total_block from ConcreteColumnIterator? @skyzh
 
Signed-off-by: ZheNing Hu <adlternative@gmail.com>

### DIFF
--- a/src/storage/secondary/column/concrete_column_iterator.rs
+++ b/src/storage/secondary/column/concrete_column_iterator.rs
@@ -28,6 +28,16 @@ pub trait BlockIteratorFactory<A: Array>: Send + Sync + 'static {
     fn get_fake_iterator(&self, index: &BlockIndex, start_pos: usize) -> Self::BlockIteratorImpl;
 }
 
+/// `ConcreteColumnIterator` Statistics
+#[derive(Debug, Default)]
+pub struct Statistics {
+    /// next_batch call times
+    next_batch_count: u32,
+
+    /// get_block call times
+    fetched_block_count: u32,
+}
+
 /// Column iterator that operates on a concrete type
 pub struct ConcreteColumnIterator<A: Array, F: BlockIteratorFactory<A>> {
     /// The [`Column`] object to iterate.
@@ -50,6 +60,9 @@ pub struct ConcreteColumnIterator<A: Array, F: BlockIteratorFactory<A>> {
 
     /// Indicate whether current_block_iter is fake.
     is_fake_iter: bool,
+
+    /// Statistics which used for reporting.
+    statistics: Statistics,
 }
 
 impl<A: Array, F: BlockIteratorFactory<A>> ConcreteColumnIterator<A, F> {
@@ -71,6 +84,10 @@ impl<A: Array, F: BlockIteratorFactory<A>> ConcreteColumnIterator<A, F> {
             finished: false,
             factory,
             is_fake_iter: false,
+            statistics: Statistics {
+                next_batch_count: 0,
+                fetched_block_count: 1,
+            },
         })
     }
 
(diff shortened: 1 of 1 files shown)


skyzh (approved):
Thanks a lot, that's awesome!

For how to get fetched blocks, every time a block is actually fetched from disk, we will call `column.get_block`. There are multiple occurrences of that:

https://github.com/risinglightdb/risinglight/blob/f1a4962da3f4626e1de57fcbf752ac362b04c643/src/storage/secondary/column/concrete_column_iterator.rs#L130

We can simply add up how many time this function is called, to see how many blocks are fetched.

For total block visited, it's simply `self.current_block_id` when the iterator is dropped. As our iterator always scans from the start to the end, `self.current_block_id` is the block number visited.

`next_batch` count is something else to sum up, and I think it's also meaningful to keep it in log.

Also I think it's better to print something like column name in the statistics, but I'm not sure whether we can get such information in storage. We may create a separate issue for this.

Do you want to implement such statistics in another PR, or continue in this PR?

skyzh:
Please also fix clippy :)

adlternative:
> Thanks a lot, that's awesome!
> 
> For how to get fetched blocks, every time a block is actually fetched from disk, we will call `column.get_block`. There are multiple occurrences of that:
> 
> https://github.com/risinglightdb/risinglight/blob/f1a4962da3f4626e1de57fcbf752ac362b04c643/src/storage/secondary/column/concrete_column_iterator.rs#L130
> 
> We can simply add up how many time this function is called, to see how many blocks are fetched.
> 
> For total block visited, it's simply `self.current_block_id` when the iterator is dropped. As our iterator always scans from the start to the end, `self.current_block_id` is the block number visited.
> 

Thanks for clarify. now the new patch have added this part.

> `next_batch` count is something else to sum up, and I think it's also meaningful to keep it in log.
> 

OK. Maybe I need more reading to learn about these parts of technical details.

> Also I think it's better to print something like column name in the statistics, but I'm not sure whether we can get such information in storage. We may create a separate issue for this.
> 
> Do you want to implement such statistics in another PR, or continue in this PR?

I look at `struct Column` and there seems to be no way to get the Column name. Maybe it's store in `struct ColumnDesc`...
I don't know if there have a way to get it.

skyzh (approved):
LGTM. Please also help revert the change of `simd.rs` so that this PR can pass CI, thanks!

PR #595, opened 2022-03-30
chore: bump toolchain and adapt to new portable simd API

Y7n05h:
Api changed in https://github.com/rust-lang/portable-simd/commit/c196b8abaafaf360647513c64fac0c8568f9b747


TennyZhuang (approved):
LGTM. It seems that our CI use incorrect toolchain, will fix it and merge this PR later.

skyzh:
CI picks the toolchain specified in `rust-toolchain`, so I think we might need to bump Rust toolchain version along with this PR...

adlternative:
LGTM. I meet this error in lastest rust-nightly.

skyzh:
> LGTM. I meet this error in lastest rust-nightly.

Hmm... Maybe there's bug with `cargo`? It should read rust toolchain of RisingLight, and force using our specified toolchain (2022-03-09).

skyzh:
Would you please provide an output of `rustup show`? If everything works correctly, there should be:

```
nightly-2022-03-09-aarch64-apple-darwin (overridden by '/Users/skyzh/Work/risinglight/rust-toolchain')
rustc 1.61.0-nightly (1eb72580d 2022-03-08)
```

adlternative:
> > LGTM. I meet this error in lastest rust-nightly.
> 
> Hmm... Maybe there's bug with `cargo`? It should read rust toolchain of RisingLight, and force using our specified toolchain (2022-03-09).

```
$ rustc --version                          
rustc 1.61.0-nightly (1d9c262ee 2022-03-26)
```
I guess that's because we use pacman in archlinux to install rust-nightly, instread of using `rustup`.

Y7n05h:
I'm very sorry. I didn't notice the version of toolchain specified in `rust-toolchain.toml` . I'm not using `rustup`, I'm just use the binaries from https://static.rust-lang.org/dist/2022-03-30 .
But I think we can make the project work with the latest rust toochain with this very simple modification. This is probably not a bad thing.

skyzh:
> I'm very sorry. I didn't notice the version of toolchain specified in rust-toolchain.toml . I'm not using rustup, I'm just use the binaries from https://static.rust-lang.org/dist/2022-03-30 .

No problem, I think we can also have some hints in https://github.com/risinglightdb/risinglight/blob/main/docs/00-develop.md to let users use cargo / rustup instead of self-setup toolchains. Meanwhile, bumping toolchain is also welcomed! Would you like to help us upgrade to toolchain 2022-03-30?

Issue #589, opened 2022-03-27
storage: support range-filter scan by sort key


skyzh:
Our storage engine already supports sorting by sort key. To leverage the property in storage, we can provide a `range_scan` interface on storage, so as to support range filter scan for executors.

To support range_scan, we should be able to start scanning from a sort key in storage engine. This requires two steps:

* record sort key in block index
* support create iterator by sort key
* add interface for executors' use

Currently, the `first_key` field is empty in block builder. We should record `first_key` in order to support seek. Note that https://github.com/risinglightdb/risinglight/issues/588 will add composite sort key support. For this issue, we may only support range filter scan by the first element in composite sort key.

https://github.com/risinglightdb/risinglight/blob/7c5e0d5193708c50b14c98f3efdddfcefbe445cd/src/storage/secondary/block/block_index_builder.rs#L39-L54

After that, we should be able to support create `RowSetIterator` by sort key.

https://github.com/risinglightdb/risinglight/blob/7c5e0d5193708c50b14c98f3efdddfcefbe445cd/src/storage/secondary/rowset/rowset_iterator.rs#L27-L33

The interface should have two new fields `start_key: Option<Vec<u8>>` and `end_key: Option<Vec<u8>>`. RowSetIterator should return rows only in this range. At the same time, the RowSetIterator should directly seek to the block of `start_key` instead of scanning from the beginning.

When everything's ready, we can finally expose range-filter scan interface to the compute layer. This requires us to modify:

https://github.com/risinglightdb/risinglight/blob/7c5e0d5193708c50b14c98f3efdddfcefbe445cd/src/storage/secondary/transaction.rs#L356-L364

PR #598, opened 2022-04-03
feat(storage): record first_key in block_index_builder

adlternative:
This patch want to complete #589 first step, which record first_key in order to support seek.

Signed-off-by: ZheNing Hu <adlternative@gmail.com>


src/storage/secondary/block/block_index_builder.rs
```diff
@@ -83,4 +87,8 @@ impl BlockIndexBuilder {
     pub fn into_index(self) -> Vec<BlockIndex> {
         self.indexes
     }
+
+    pub fn set_first_key(&mut self, first_key: Vec<u8>) {
```
skyzh, line 91: I'd prefer having the signature as

```
pub fn set_first_key(&mut self, first_key: &[u8]) {
  self.first_key.clear();
  self.first_key.extend(first_key);
}
```

So as to reduce allocation as much as possible.

By the way, it seems that `first_key` will only be used on `finish_block`. Is it possible to make it a parameter of `finish_block` instead of having a separate function?

src/storage/secondary/column/primitive_column_builder.rs
```diff
@@ -121,7 +121,7 @@ pub fn append_one_by_one<'a, A: Array>(
 impl<T: PrimitiveFixedWidthEncode> ColumnBuilder<T::ArrayType> for PrimitiveColumnBuilder<T> {
     fn append(&mut self, array: &T::ArrayType) {
         let mut iter = array.iter().peekable();
-
+        let mut first_flag = true;
```
skyzh, line 124: first_flag should be an inner state of `ColumnBuilder` instead of a local variable. One append doesn't necessarily produce a block.

I think a better way is to store the first_key in the struct on switching builder (when `self.current_builder.is_none()`)?

skyzh (commented):
By the way, first_key is only useful when the column is fully sorted. So we can have a parameter on column builder whether to store the first_key or not.

src/storage/secondary/block/block_index_builder.rs
adlternative: > By the way, first_key is only useful when the column is fully sorted. So we can have a parameter on column builder whether to store the first_key or not.

Agree. But I am curious about where can we know that column need sort or not? Maybe pass from `ColumnBuilderOptions` later?

adlternative: > I'd prefer having the signature as
> 
> ```
> pub fn set_first_key(&mut self, first_key: &[u8]) {
>   self.first_key.clear();
>   self.first_key.extend(first_key);
> }
> ```
> 
> So as to reduce allocation as much as possible.
> 
> By the way, it seems that `first_key` will only be used on `finish_block`. Is it possible to make it a parameter of `finish_block` instead of having a separate function?

This means that we need to modify `PrimitiveColumnBuilder<T>::finish_builder()` interface, that's ok...

src/storage/secondary/column/primitive_column_builder.rs
adlternative: > first_flag should be an inner state of `ColumnBuilder` instead of a local variable. One append doesn't necessarily produce a block.
> 
> I think a better way is to store the first_key in the struct on switching builder (when `self.current_builder.is_none()`)?

Yes, the condition `self.current_builder.is_none()` can be reused.

adlternative:
By the way, should we apply this feature to `CharColumnBuilder` or `BlobColumnBuilder`, they have similar interface...

skyzh:
> By the way, should we apply this feature to `CharColumnBuilder` or `BlobColumnBuilder`, they have similar interface...

Yes, indeed. But we can work on primitive types first, and add support for other columns later.

adlternative:
Hi, now we can use a new option `record_first_key` to control if we want record first key of the block, is this change reasonable?
I am not sure about if named it as `is_sorted` or something else will be better?
@skyzh

skyzh:
> Hi, now we can use a new option record_first_key to control if record first key of the block, is this change reasonable?
I am not sure about if named it as is_sorted or something else will be better?

I think `record_first_key` is clear enough! I'll take a look at this PR soon. Thanks!

src/storage/secondary/block/block_index_builder.rs
```diff
@@ -42,14 +42,15 @@ impl BlockIndexBuilder {
...
+        first_key: &mut Vec<u8>,
     ) {
         self.indexes.push(BlockIndex {
             offset: column_data.len() as u64,
             length: block_data.len() as u64 + BLOCK_HEADER_SIZE as u64,
             first_rowid: self.last_row_count as u32,
             row_count: (self.row_count - self.last_row_count) as u32,
             /// TODO(chi): support sort key
-            first_key: "".into(),
+            first_key: first_key.drain(..).collect(),
```
skyzh, line 53: I'd prefer to get a full ownership of `first_key` in this function. Otherwise this interface would look confusing 🤣

src/storage/secondary/column/blob_column_builder.rs
```diff
@@ -56,8 +60,13 @@ impl BlobColumnBuilder {
...
         };
 
-        self.block_index_builder
-            .finish_block(block_type, &mut self.data, &mut block_data, stats);
+        self.block_index_builder.finish_block(
+            block_type,
+            &mut self.data,
+            &mut block_data,
+            stats,
+            &mut self.first_key,
```
skyzh, line 68: Let's drain here: self.first_key.drain(..).collect_vec()

src/storage/secondary/column/blob_column_builder.rs
```diff
@@ -80,6 +89,11 @@ impl ColumnBuilder<BlobArray> for BlobColumnBuilder {
                         PlainBlobBlockBuilder::new(self.options.target_block_size - 16),
                     ));
                 }
+                if let Some(to_be_appended) = iter.peek() {
+                    if self.options.record_first_key && to_be_appended.is_some() {
```
skyzh, line 93: `None` is also a valid first_key. I think we can decide on encoding of None keys.

Maybe we can use `\0` to represent null keys, and `\1` + BlobEncode to represent non-null keys?

skyzh: Also for the encoding of the key, we can use utilities provided in https://github.com/risinglightdb/risinglight/blob/main/src/storage/secondary/encode.rs

For blob types, simply call `to_be_appended.to_byte_slice().to_vec()`, and store it in `first_key`.

skyzh (commented):
Generally LGTM, but there're still some misc fixes to do. Thanks for your contribution!

src/storage/secondary/column/blob_column_builder.rs
skyzh: We can also `assert!(self.first_key.is_empty())` here, for an extra layer of sanity check.

skyzh:
By the way, please have some basic unit tests, at least for primitive type builders 🥰

src/storage/secondary/column/blob_column_builder.rs
adlternative: > `None` is also a valid first_key. I think we can decide on encoding of None keys.
> 
> Maybe we can use `\0` to represent null keys, and `\1` + BlobEncode to represent non-null keys?

So we use `\1` for `CharColumnBuilder` too, right?

skyzh: Yes, we use this as a temporary encoding for nullable elements, for both char, blob and primitive types.

adlternative:
> By the way, please have some basic unit tests, at least for primitive type builders smiling_face_with_three_hearts

Hi, I have followed your suggestions to add `is_first_key_null` to proto and add some tests for primitive type builders. 😉
@skyzh

src/storage/secondary/column/char_column_builder.rs
```diff
@@ -130,6 +144,13 @@ impl ColumnBuilder<Utf8Array> for CharColumnBuilder {
                         nullable
                     ),
                 }
+
+                if let Some(to_be_appended) = iter.peek() {
+                    if self.options.record_first_key && to_be_appended.is_some() {
+                        self.is_first_key_null = false;
+                        self.first_key = to_be_appended.unwrap().as_bytes().to_vec();
+                    }
```
skyzh, line 152: It seems that once `is_first_key_null` is set to `false`, it will never be set back to `true`?

We can define `self.first_key` as `Option<Vec<u8>>`, so as to make the logic clearer.

```rust
if self.options.record_first_key {
   self.first_key = to_be_appended.map(|x| x.as_bytes().to_vec());
}
```

And we can modify `finish_block` to accept `first_key: Option<Vec<u8>>`. Inside the block builder, we match `first_key`. If null, set `is_first_key_null` to true. Otherwise, set `first_key` to the inner value.

skyzh (commented):
Still some small issues. Thanks again for your contribution!

src/storage/secondary/column/char_column_builder.rs
adlternative: > Still some small issues. Thanks again for your contribution!

Thanks also for your review too! 😭

adlternative: > It seems that once `is_first_key_null` is set to `false`, it will never be set back to `true`?
> 
> We can define `self.first_key` as `Option<Vec<u8>>`, so as to make the logic clearer.
> 
> ```rust
> if self.options.record_first_key {
>    self.first_key = to_be_appended.map(|x| x.as_bytes().to_vec());
> }
> ```
> 
> And we can modify `finish_block` to accept `first_key: Option<Vec<u8>>`. Inside the block builder, we match `first_key`. If null, set `is_first_key_null` to true. Otherwise, set `first_key` to the inner value.

Maybe `finish_block()` will be like this?

```rust
    pub fn finish_block(
        &mut self,
        block_type: BlockType,
        column_data: &mut Vec<u8>,
        block_data: &mut Vec<u8>,
        stats: Vec<BlockStatistics>,
        first_key: &mut Option<Vec<u8>>,
    ) {
        self.indexes.push(BlockIndex {
            offset: column_data.len() as u64,
            length: block_data.len() as u64 + BLOCK_HEADER_SIZE as u64,
            first_rowid: self.last_row_count as u32,
            row_count: (self.row_count - self.last_row_count) as u32,
            /// TODO(chi): support sort key
            first_key: match first_key {
                Some(x) => x.drain(..).collect(),
                None => Default::default(),
            },
            stats,
            is_first_key_null: first_key.is_none(),
        });
        ...
```

skyzh: Exactly! And I’d prefer to have Option<Vec<u8>> as parameter instead of &mut. As block is typically large, this re-allocation is acceptable.

skyzh (approved):
LGTM, good work!

adlternative:
> LGTM, good work!

Thanks a lot! @skyzh

Issue #592, opened 2022-03-27
executor: optimize `NestedLoopJoin`'s memory usage


skyzh:
https://github.com/risinglightdb/risinglight/blob/7c5e0d5193708c50b14c98f3efdddfcefbe445cd/src/executor/nested_loop_join.rs#L26-L32

Currently, NLJ will pull all data into memory before start processing data. This would make memory usage high.

Optimally, NLJ should work like...

```
for chunk in left {
  for chunk in right {
    // process chunk
  }
}
```

However, this requires us to support restarting a child executor (restart the right child), which is not possible in our system. Therefore, for the first step, we'd like to optimize NLJ like...

```
let bufferred_left = pull all data from left side
for chunk in right {
  for chunk in bufferred_left {
    // process chunk
  }
}
```

BaymaxHWY:
hello, may i try this?

BaymaxHWY:
I have a question about this. Now, NLJ is waiting for enough row(one DataChunk) to eval and filter(this is deferred),
but when one right_chunk is obtained from right_child, it must be processed immediately(**right outer join**).

skyzh:
> I have a question about this. Now, NLJ is waiting for enough row(one DataChunk) to eval and filter(this is deferred),
> but when one right_chunk is obtained from right_child, it must be processed immediately(**right outer join**).
> Maybe should hold a HashMap to record the right_row's state(Whether there is a match left_row)?

I think for the first step, we can choose to support join of only one direction (either left join or right join)? The optimizer can rewrite left join to a right join, and back forth.

BaymaxHWY:
> we can choose to support join of only one direction (either left join or right join)

this is easy work, but maybe the optimizer is hard work( how to support full outer join and multi-table join)? 
Is it necessary to complete the optimizer part before continuing this issue or synchronization?

skyzh:
> how to support full outer join and multi-table join

I think we can have some regression in tests (maybe we need to disable some of the test cases in https://github.com/risinglightdb/risinglight/blob/main/tests/sql/join.slt)? Just took a look at the current plan, it seems that for outer join and right outer join, optimizer will always choose NLJ. Therefore, we can disable some test cases in `join.slt`.

> Is it necessary to complete the optimizer part before continuing this issue or synchronization?

[...]

PR #599, opened 2022-04-03
feat(executor):optimize NestedLoopJoin's memory usage

BaymaxHWY:
about issue: [#598](https://github.com/risinglightdb/risinglight/issues/592)
### Change
1. NLJ only support inner and left outer join
2. disable the test of `join.slt`
3. add sql test file `issue_592.slt`, it just test inner join and left join


tests/sqllogictest/build.rs
```diff
@@ -9,8 +9,8 @@ fn main() {
     println!("cargo:rerun-if-changed=../../tests/sql");
 
     const PATTERN: &str = "../../tests/sql/**/[!_]*.slt"; // ignore files start with '_'
-    const MEM_BLOCKLIST: &[&str] = &["statistics.slt"];
-    const DISK_BLOCKLIST: &[&str] = &[];
+    const MEM_BLOCKLIST: &[&str] = &["statistics.slt", "join.slt"];
```
skyzh, line 12: If a test file is ignored for both cases, consider renaming it to `_join.slt` instead of changing the blocklist. Therefore, developers will be sure that it is disabled.

src/executor/nested_loop_join.rs
```diff
@@ -22,40 +22,42 @@ pub struct NestedLoopJoinExecutor {
...
     pub async fn execute(self) {
-        // collect all chunks from children
-        let (left_chunks, right_chunks) = async {
-            tokio::try_join!(
-                self.left_child.try_collect::<Vec<DataChunk>>(),
-                self.right_child.try_collect::<Vec<DataChunk>>(),
-            )
-        }
-        .await?;
+        // only support inner and left outer join
```
skyzh, line 25: We should also panic on `BoundJoinOperator::RightOuter | BoundJoinOperator::FullOuter` to avoid producing wrong result.

src/executor/nested_loop_join.rs
```diff
@@ -77,12 +79,12 @@ impl NestedLoopJoinExecutor {
             self.join_op,
             BoundJoinOperator::LeftOuter | BoundJoinOperator::FullOuter
```
skyzh, line 80: remove FullOuter here?

src/executor/nested_loop_join.rs
```diff
@@ -77,12 +79,12 @@ impl NestedLoopJoinExecutor {
...
-            let mut i = 0;
-            for left_row in left_rows() {
+            let left_row_num = left_rows().count();
+            for (mut i, left_row) in left_rows().enumerate() {
                 let mut matched = false;
-                for _ in right_rows() {
+                for _ in 0..right_row_num {
                     matched |= matches!(filter.get(i), Some(true));
-                    i += 1;
+                    i += left_row_num;
```
skyzh, line 87: After reading this part 3 times, I'm finally sure that's correct... Would you please add some comments on what this part is doing? For example, having something like:

```rust
// Now the filter map contains the selected rows of the cross join result `right x left`. We will need to figure out whether there's an unmatched left row, and produce the correct outer join result.
// To compute the unmatched rows, we will need to first pick a row from left (namely the i-th row). Then we check if all `filter[i + left_row_num * j]` are false. If all false, we produce a (left, NULL) row.
```

Please add such comments across L82-L93.

skyzh (commented):
Rest LGTM, good work!

tests/sqllogictest/build.rs
BaymaxHWY: Get it!

skyzh (approved):
LGTM, thanks!

Issue #597, opened 2022-04-02
executor: implement `SortMergeJoin`


skyzh:
https://en.wikipedia.org/wiki/Sort-merge_join

D2Lark:
I'd like to try
whether the children of SortMergeJoin are ordered？

skyzh:
Yes, they can assumed to be ordered. We can ensure this in optimizer.

PR #600, opened 2022-04-03
feat(executor): implement `SortMergeJoin` executor

D2Lark:
close https://github.com/risinglightdb/risinglight/issues/597
Should I add this executor to the PlanVisitor trait?  I have marked this code with #[allow(dead_code)] to pass Clippy test


skyzh:
> Should I add this executor to the PlanVisitor trait?

I think we can do this in another PR, if you want to work on the optimizer and actually make our system support `SortMergeJoin`! You may create a separate issue for that, and work on it (or wait for another developer to pick it up).

> I have marked this code with #[allow(dead_code)] to pass Clippy test

I agree on that! Furthermore, I suggest adding `#![allow(dead_code)]` at the top of `sort_merge_join` module, so that we won't need to write it anywhere. We can remove it after we support SortMergeJoin in optimizer.

Please also add some unit tests for sort merge join, and ping me when you are ready. Thanks for your contribution!

src/executor/sort_merge_join.rs
```diff
@@ -0,0 +1,96 @@
...
+        let (left_chunks, right_chunks) = async {
+            tokio::try_join!(
+                self.left_child.try_collect::<Vec<DataChunk>>(),
+                self.right_child.try_collect::<Vec<DataChunk>>(),
+            )
+        }
+        .await?;
+
+        // get rows iterator from chunks
+        let left_rows = || left_chunks.iter().flat_map(|chunk| chunk.rows()).peekable();
```
D2Lark, line 36: What is the purpose of this closure? If I remove this closure and return the iterator trait directly .
    #[try_stream(boxed, ok = DataChunk, error = ExecutorError)] will throw this error 
`risinglight::executor
pub enum ExecutorError
The error type of execution.

[7 implementations]()

mismatched types
expected opaque type `impl Iterator<Item = data_chunk::RowRef<'_>>`
   found opaque type `impl Iterator<Item = data_chunk::RowRef<'_>>`rustc[E0308](https://doc.rust-lang.org/error-index.html#E0308)
[data_chunk.rs(60, 27): ]()the expected opaque type
[data_chunk.rs(60, 27): ]()the found opaque type
implementation of `FnOnce` is not general enough
closure with signature `fn(&'0 data_chunk::DataChunk) -> impl Iterator<Item = data_chunk::RowRef<'_>>` must implement `FnOnce<(&data_chunk::DataChunk,)>`, for any lifetime `'0`...
...but it actually implements `FnOnce<(&data_chunk::DataChunk,)>`

src/executor/sort_merge_join.rs
```diff
@@ -0,0 +1,96 @@
...
+            right_chunks
+                .iter()
+                .flat_map(|chunk| chunk.rows())
+                .peekable()
+        };
+
+        // build
+        let data_types = self.left_types.iter().chain(self.right_types.iter());
+        let mut builder = DataChunkBuilder::new(data_types, PROCESSING_WINDOW_SIZE);
+        let mut left_row = left_rows().next().unwrap();
```
D2Lark, line 47: I find 'left_rows().next().unwrap' didn't return the next row in rows. Is there a method other than for row in left_rows() can retun next row ? I found for method in [here](https://github.com/risinglightdb/risinglight/blob/a23b9b15337ff4aedcbbec9d638acb0e0562822e/src/executor/hash_join.rs#L56)

src/executor/sort_merge_join.rs
skyzh: The closure is used to flatten data chunk into rows with minimal overhead.

`left_chunks` is `Vec<DataChunk>`. We convert an iterator on data chunk to an iterator of `RowRef<'_>` using this function. There's no data clone within the process.

And your problem looks like a lifetime problem. Not sure what have you done, but it seems that the function signature is only valid for a single lifetime, instead of for all lifetimes.

By the way, if possible, I'd also prefer not to fetch all rows in advance. SortMergeJoin should be able to go over the data chunks from the child executors one by one.

src/executor/sort_merge_join.rs
skyzh: left_rows() is an iterator. You may store it in advance.

```rust
let left_rows_iter = left_rows();
let first_row = left_rows_iter.next().unwrap();
```

By the way, unwrapping here is dangerous, as there might be no more rows.

src/executor/sort_merge_join.rs
D2Lark: > By the way, if possible, I'd also prefer not to fetch all rows in advance. SortMergeJoin should be able to go over the data chunks from the child executors one by one.
I understand what you mean, loading all the chunks at once is not that necessary, I'll optimize it again

src/executor/sort_merge_join.rs
D2Lark: > By the way, unwrapping here is dangerous, as there might be no more rows.
Thanks for reminding me！I should be more careful

src/executor/sort_merge_join.rs
```diff
@@ -0,0 +1,149 @@
...
+    left_column_indexes: &[usize],
+    right_column_indexes: &[usize],
+) -> Ordering {
+    let left_data_value = left_row.get_by_indexes(left_column_indexes)[0].clone();
+    let right_data_value = right_row.get_by_indexes(right_column_indexes)[0].clone();
+    left_data_value.partial_cmp(&right_data_value).unwrap()
+}
+
+#[try_stream(boxed,ok = RowRef<'static>, error = ExecutorError)]
+async fn row_stream(result: Result<DataChunk, ExecutorError>) {
```
skyzh, line 102: That's one of the most weird (and unsolved) problem of the Rust programming language -- you simply cannot return something with lifetime from a stream of fully owned data.

I'd recommend you to write in the following way:

```rust
#[try_stream(boxed, ok = Row, error = ExecutorError)]
async fn row_stream(stream: BoxedExecutor) {
    #[for_await]
    for chunk in stream {
        for row in chunk?.rows() {
            yield row.values().collect();
        }
    }
}
```

Note that this is not as efficient as before, as now we have to create a copy of each row when yielding (`row.values().collect()` creates a fully owned Row). But we can optimize it later.

We can keep everything simple in this PR, and see how to optimize things later, if we want to use `RowRef` and do zero extra allocations during the whole process.

D2Lark: Thanks for such a nice reply, I just thought it was a simple lifetime problem, but I didn't expect it to be so complicated... Anyway, I'll finish this simple PR as soon as I can and come back with a better way to optimize it.

src/executor/sort_merge_join.rs
```diff
@@ -0,0 +1,147 @@
...
+            let mut builder = DataChunkBuilder::new(data_types, PROCESSING_WINDOW_SIZE);
+
+            loop {
+                match compare_row(
+                    &left_row,
+                    &right_row,
+                    self.left_column_index,
+                    self.right_column_index,
+                ) {
+                    Ordering::Equal => {
```
skyzh, line 43: Still some little problem with `Equal`. Assuming you are doing an inner join. If we have:

left: [1, 1]
right: [1, 1]

The join result should be four lines of [1, 1]. However, the current implementation doesn't seem correct?

Please also add this test case, thanks!

src/executor/sort_merge_join.rs
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+    }
+}
+
+pub fn compare_row(
+    left_row: &Row,
+    right_row: &Row,
+    left_column_indexes: usize,
+    right_column_indexes: usize,
```
xxchan, line 82: Are they `index` or `indexes`?

src/executor/sort_merge_join.rs
```diff
@@ -0,0 +1,147 @@
...
+        let mut right_rows = row_stream(self.right_child);
+        let mut left_row;
+        let mut right_row;
+        let first_left_row = left_rows.next().await;
+        let first_right_row = right_rows.next().await;
+        if first_left_row.is_none() || first_right_row.is_none() {
+            yield Err(ExecutorError::NotNullable)?
+        } else {
+            left_row = first_left_row.unwrap()?;
+            right_row = first_right_row.unwrap()?;
```
xxchan, line 31: A style suggestion: 
```rust
match (..., ...) {
    Some(mut left_row), Some(mut right_row) => {...}
    _ => {...}
}
```
This might be more clear than `is_none` and `unwrap`?

src/executor/sort_merge_join.rs
D2Lark: > The join result should be four lines of [1, 1]. However, the current implementation doesn't seem correct?

I found that the bug you mentioned was caused by the unit tests not being constructed correctly.
So I decided to make DataChunk contain two columns and added some test cases, and now it's okay.

src/executor/sort_merge_join.rs
D2Lark: Thanks, I've changed it once to put the logic of is_none in front, but your solution is clearer

src/executor/sort_merge_join.rs
```diff
@@ -0,0 +1,146 @@
...
+#[cfg(test)]
+mod tests {
+    use test_case::test_case;
+
+    use super::*;
+    use crate::array::ArrayImpl;
+    use crate::types::{DataTypeExt, DataTypeKind};
+    #[test_case(vec![1],vec![1],vec![1])]
+    #[test_case(vec![1],vec![1,2,3],vec![1])]
+    #[test_case(vec![1,2,3,4],vec![2,4,6],vec![2,4])]
```
xxchan, line 108: I'm not sure whether I understand correctly, but maybe add the case mentioned above?

```suggestion
    #[test_case(vec![1],vec![1],vec![1])]
    #[test_case(vec![1],vec![1,2,3],vec![1])]
    #[test_case(vec![1,2,3,4],vec![2,4,6],vec![2,4])]
    #[test_case(vec![1,1],vec![1,1],vec![1,1,1,1])]
```

skyzh:
The join implementation still looks a little bit suspicious to me -- we didn't consider duplicated elements.

Consider this case:

```
#[test_case(vec![1,1,1],vec![1,1,1],vec![1,1,1,1,1,1,1,1,1])]
```

The result should be `3 * 3 = 9` number of 1 s, according to the definition of inner join.

Probably we will need some special handling when two columns equal. For example,

```
Equal => {
  let mut left_equal_cnt = 1;
  while let Some(row) = left_rows.next().await {
    if row == current_left_row { left_equal_cnt += 1; } else { break; }
  }
  let mut right_equal_cnt = 1;
  while let Some(row) = left_rows.next().await {
    if row == current_left_row { left_equal_cnt += 1; } else { break; }
  }
  // push to build `left_equal_cnt * right_equal_cnt` times.
}
```

This algorithm is only correct when doing equal inner join (e.g., select * from a inner join b on a.x = b.x, where a.x=b.x is the only equal condition).

We can just implement like above, and add more join types in the future.

src/executor/sort_merge_join.rs
skyzh: Let's support single index condition first. We can later support multiple join keys, or even non-equal conditions.

skyzh: ... and in this case, better to rename to `index`.

D2Lark:
> Probably we will need some special handling when two columns are equal. For example,

I realized that I forgot to consider a situation in which there are lots of the same join keys in one column.
Refer [Tidb implement](https://pingcap.com/zh/blog/tidb-source-code-reading-15)
>顺序读取外表 A 直到 join-keys 中出现另外的值，把相同 **keys** 的行放入数组 a1，同样的规则读取内表 B，把相同 **keys** 的行放入数组 a2。如果外表数据或者内表数据读取结束，退出。

if we have:
left chunk
1 a
1 b 
2 c
2 d

right chunk
2 e
2 f

result_chunk
2 c 2 e
2 c 2 f
2 d 2 e
2 d 2 f

emm... This is equivalent to doing a full join on the datachunks with the same keys

test cases that  I should consider：

- 1.child_row_len > or < PROCESSING_WINDOW_SIZE
- 2.duplicate element in  join column

D2Lark:
> This algorithm is only correct when doing equal inner join (e.g., select * from a inner join b on a.x = b.x, where a.x=b.x is the only equal condition).
> 
> We can just implement like above, and add more join types in the future.

I found this comment in tidb source code

> // MergeJoinExec implements the merge join algorithm.
> // This operator assumes that two iterators of both sides
> // will provide required order on join condition:
> // 1. For equal-join, one of the join key from each side
> // matches the order given.
> // 2. For other cases its preferred not to use SMJ and operator
> // will throw error.

SortMergJoin seems to only  fit one condition such as a.v1 == b.v2 .I don't think SMJ applies to inner join scenarios with multiple join keys...

skyzh:
> SortMergJoin seems to only fit one condition such as a.v1 == b.v2 .I don't think SortMergJoin applies to inner join scenarios with multiple join keys...

Okay, we can later only enable sort merge join on inner join && only one key in equal condition. Anyway, it's necessary to fix the case for duplicated rows 🤪

D2Lark:
> Okay, we can later only enable sort merge join on inner join && only one key in equal condition. Anyway, it's necessary to fix the case for duplicated rows 🤪

Now test cases seem work after I convert same key row to a chunk .
just a liter problem from Clippy , I don't know how to mess with this

> warning: unsequenced read of `__task_context`
>   --> src/executor/sort_merge_join.rs:19:5
>    |
> 19 |     #[try_stream(boxed, ok = DataChunk, error = ExecutorError)]
>    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>    |
>    = note: `#[warn(clippy::eval_order_dependence)]` on by default

src/executor/sort_merge_join.rs
```diff
@@ -0,0 +1,217 @@
...
+                            }
+                        }
+                    }
+                }
+                // if rows line < PROCESSING_WINDOW_SIZE ,take rest rows out of builder
+                if let Some(chunk) = { builder }.take() {
+                    yield chunk;
+                }
+            }
+            _ => yield Err(ExecutorError::NotNullable)?,
```
skyzh, line 78: If any side is empty, maybe we should just break instead of returning an error? It is okay to join a table with empty table, and the result (if inner join) is an empty table.

src/executor/sort_merge_join.rs
```diff
@@ -0,0 +1,217 @@
...
+                    chunk.push(row);
+                } else {
+                    yield chunk;
+                    chunk = Vec::new();
+                    current_row = Some(row);
+                    chunk.push(current_row.clone().unwrap());
+                }
+            }
+            Err(e) => {
+                yield Err(e)?;
```
skyzh, line 132: return e;

src/executor/sort_merge_join.rs
```diff
@@ -0,0 +1,217 @@
...
+            }
+        }
+    }
+    if !chunk.is_empty() {
+        yield chunk;
+    } else {
+        yield Err(ExecutorError::NotNullable)?;
```
skyzh, line 139: No need to error, just return?

skyzh (commented):
Rest LGTM, good work!

skyzh:
> just a liter problem from Clippy , I don't know how to mess with this

Seems some issue with try_stream. I suspect that's because `yield Err(e)?`. Change it to `return Err(e)` might help fix. If not, just ignore the lint.

src/executor/sort_merge_join.rs
skyzh: Or what about simply `let row = row?`

D2Lark: This looks good！

src/executor/sort_merge_join.rs
D2Lark: I change return chunk directly `    yield chunk; ` Empty chunk is acceptable

src/executor/sort_merge_join.rs
D2Lark: how do I return an empty DataChunk when one of child is empty?

skyzh: No need to yield anything, do nothing.

You may re-write this part as follows:

```rust
        let left = left_chunks.next().await;
        let right = right_chunks.next().await;
        if let (Some(left_chunk), Some(right_chunk)) = (left, right) {
        }
```

This will also solve the clippy's evaluation order warning.

skyzh (commented):
Looks like the unit test failed? 🤣 Please help fix before this PR can get merged. Thanks!

D2Lark:
> Looks like the unit test failed? 🤣 Please help fix before this PR can get merged. Thanks!

🤣 What if this is the case? ` a =  [1,2,3] b = [4,5,6] `then nothing be yield?

skyzh:
> 



> > Looks like the unit test failed? 🤣 Please help fix before this PR can get merged. Thanks!
> 
> 🤣 What if this is the case? `a = [1,2,3] b = [4,5,6]`then nothing be yield?

Yes, should yield nothing.

skyzh (approved):
I see, you'll need to change the test case to handle empty output correctly. As the implementation is correct, you may send a separate PR to enable this test case.

skyzh:
... and I think we can have a utility function like `collect_from_executor`, which collects all output from executor and make them into a single chunk. This will make your life a lot easier.

D2Lark:
> ink we can have a utility function like `collect_from_executor`, which collects all o

I will try this.

PR #603, opened 2022-04-04
chore: bump to v0.1.3

skyzh:


TennyZhuang (approved):

skyzh:
Seems that prost 0.10 is causing significant compile time regression?

skyzh:
I believe the regression is caused by https://github.com/tokio-rs/prost/pull/610, maybe we should stay on 0.9...

Issue #611, opened 2022-04-07
executor: introduce DataChunkBuilder and split chunks in LimitExecutor&OrderExecutor


wangqiim:
introduce DataChunkBuilder and split chunks in LimitExecutor&OrderExecutor after #610

wangqiim:
I will deal with it

PR #612, opened 2022-04-07
feat(executor): introduce DataChunkBuilder and split chunks in Order …

wangqiim:
…and Limit
issue #611 
1. Fix panic bug and add test cases. 
```sql
select * 
from t 
limit 0 offset 3
```
2. Don't modify average_agg return type ~~when aggregated coloum datatype is Decimal~~.
3. Introduce DataChunkBuilder and split chunks in Order ~~and Limit~~.

Signed-off-by: Qi Wang <wangqiim@163.com>

### DIFF
--- a/src/binder/expression/agg_call.rs
+++ b/src/binder/expression/agg_call.rs
@@ -89,16 +89,7 @@ impl Binder {
             }
         }
         let (kind, return_type) = match func.name.to_string().to_lowercase().as_str() {
-            "avg" => {
-                let agg_kind = AggKind::Avg;
-                let mut default_data_type = Some(DataType::new(DataTypeKind::Double, false));
-                if let Some(ref data_type) = args[0].return_type() {
-                    if let DataTypeKind::Decimal(_, _) = data_type.kind {
-                        default_data_type = args[0].return_type();
-                    }
-                }
-                (agg_kind, default_data_type)
-            }
+            "avg" => (AggKind::Avg, args[0].return_type()),
             "count" => {
                 if args.is_empty() {
                     let first_index_column = BoundExpr::InputRef(BoundInputRef {
--- a/src/binder/expression/agg_call.rs
+++ b/src/binder/expression/agg_call.rs
@@ -89,10 +89,16 @@ impl Binder {
             }
         }
         let (kind, return_type) = match func.name.to_string().to_lowercase().as_str() {
-            "avg" => (
-                AggKind::Avg,
-                Some(DataType::new(DataTypeKind::Double, false)),
-            ),
+            "avg" => {
+                let agg_kind = AggKind::Avg;
+                let mut default_data_type = Some(DataType::new(DataTypeKind::Double, false));
+                if let Some(ref data_type) = args[0].return_type() {
+                    if let DataTypeKind::Decimal(_, _) = data_type.kind {
+                        default_data_type = args[0].return_type();
+                    }
+                }
+                (agg_kind, default_data_type)
+            }
             "count" => {
                 if args.is_empty() {
                     let first_index_column = BoundExpr::InputRef(BoundInputRef {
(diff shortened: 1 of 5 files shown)


src/binder/expression/agg_call.rs
```diff
@@ -89,10 +89,16 @@ impl Binder {
...
+            "avg" => {
+                let agg_kind = AggKind::Avg;
+                let mut default_data_type = Some(DataType::new(DataTypeKind::Double, false));
+                if let Some(ref data_type) = args[0].return_type() {
+                    if let DataTypeKind::Decimal(_, _) = data_type.kind {
+                        default_data_type = args[0].return_type();
+                    }
+                }
+                (agg_kind, default_data_type)
+            }
```
xxchan, line 101: Could you explain a bit why to change this (in this PR)? It seems not relevant. 🤔

src/executor/limit.rs
```diff
@@ -1,52 +1,49 @@
...
                 continue;
             }
-            if (start..end) == (0..cardinality) {
-                yield batch;
-            } else {
-                yield batch.slice(start..end);
+            for row in batch.rows().skip(start).take(end - start) {
+                if let Some(chunk) = builder.push_row(row.values()) {
+                    yield chunk
+                }
```
xxchan, line 38: Iterating over batch here seems to introduce addtional overheads. Maybe we just don't need to split chunks in limit?

xxchan: I think some executors need to split chunks because they iterate over the inputs row by row and then generate different distribution of data, e.g., sort, join. The inputs are splitted, but the output is a single big chunk.

For limit, it only forwards input chunks, so no addtional effort to split chunks is needed.

src/binder/expression/agg_call.rs
wangqiim: For example in q1.sql.
```sql
select
    l_returnflag,
    l_linestatus,
    sum(l_quantity) as sum_qty,
    sum(l_extendedprice) as sum_base_price,
    sum(l_extendedprice * (1 - l_discount)) as sum_disc_price,
    sum(l_extendedprice * (1 - l_discount) * (1 + l_tax)) as sum_charge,
    avg(l_quantity) as avg_qty,
    avg(l_extendedprice) as avg_price,
    avg(l_discount) as avg_disc,
    count(*) as count_order
from
    lineitem
where
    l_shipdate <= date '1998-12-01' - interval '71' day
group by
    l_returnflag,
    l_linestatus
order by
    l_returnflag,
    l_linestatus;
```
For `avg(l_quantity) as avg_qty`、`avg(l_extendedprice) as avg_price`、`avg(l_discount) as avg_disc`，the evaluate result is `DataTypeKind::Decimal`, But The output_type of excuter will be `DataTypeKind::Double`. So it will panic when DataChunkBuilder.push_row(row.values) in `order.rs`.

xxchan: I think the problem arose because when we rewrite `avg` into `sum / count` (a few lines below, starting from the `match kind`), the `return_type` mismatches `left_expr`'s type. Would it be better to fix the problem there?

src/binder/expression/agg_call.rs
```diff
@@ -89,10 +89,7 @@ impl Binder {
             }
         }
         let (kind, return_type) = match func.name.to_string().to_lowercase().as_str() {
-            "avg" => (
-                AggKind::Avg,
-                Some(DataType::new(DataTypeKind::Double, false)),
-            ),
```
xxchan, line 95: I'm a little bit unsure and may need some time to look into the details of how the type system work. 🤔

xxchan (commented):
Others LGTM

src/binder/expression/agg_call.rs
skyzh: Avg will be rewritten, so I think it's okay to fill any type here.

skyzh (approved):

PR #616, opened 2022-04-10
chore(test): add more test case

D2Lark:
Signed-off-by: d2lark <lichengamoy@gmail.com>

complete test case for sort-merge-join exexcutor

### DIFF
--- a/src/executor/sort_merge_join.rs
+++ b/src/executor/sort_merge_join.rs
@@ -158,17 +158,10 @@ fn full_join(left_chunk: &Vec<Row>, right_chunk: &Vec<Row>) -> Vec<Row> {
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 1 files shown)


skyzh (approved):
LGTM!

By the way, personally I don't quite like the `test_case` macro. If we click on `run test` in VSCode, these test case won't be picked up. They have a strange naming convention.

I'd prefer to have multiple functions calling the test helper. For example:

```
#[tokio::test]
async fn test_duplicated_elements() {
  test_helper(vec![1, 1], vec![1, 1], vec![1,1,1,1]).await;
}
```

Do you want to get this changed in this PR?

D2Lark:
> Do you want to get this changed in this PR?

I also think the test case bug is a bit of a bummer, it's better written this way

Issue #588, opened 2022-03-27
storage: support composite sort key in MergeIterator


skyzh:
Currently, `MergeIterator` only supports taking a single `sort_idx` as parameter:

https://github.com/risinglightdb/risinglight/blob/7c5e0d5193708c50b14c98f3efdddfcefbe445cd/src/storage/secondary/merge_iterator.rs#L34-L36

We should make `MergeIterator` be able to accept a vector of `sort_idx`. For example, `sort_key_idx = [2, 1, 0]`:

* Implies that the child iterators are already being sorted using 3 columns: firstly be column `#2`, then by column `#1`, at last by `#0`.
*MergeIterator should also sort by this rule.

```rust
pub fn new(iters: Vec<SecondaryIterator>, sort_key_idx: Vec<usize>) -> Self {
```

After we finish the refactor in the storage layer, we can firstly make all places to pass `vec![sort_key]` instead of `sort_key`. Then, we can refactor little by little to support full composite sort key support across our system.

BaymaxHWY:
Hi, can i try this? This requirement just changes the code about `sort_key_idx` and the function `compare_data()`?

PR #622, opened 2022-04-12
feat(storage): add support composite sort key in MergeIterator

BaymaxHWY:
close issue: [#588 ](https://github.com/risinglightdb/risinglight/issues/588)

### DIFF
--- a/src/storage/secondary/merge_iterator.rs
+++ b/src/storage/secondary/merge_iterator.rs
@@ -50,25 +50,19 @@ impl MergeIterator {
         (left_id, left_batch_row_id): &(usize, usize),
         (right_id, right_batch_row_id): &(usize, usize),
     ) -> std::cmp::Ordering {
-        let mut res = Ordering::Equal;
+        let left_chunk = self.chunk_buffer[*left_id].as_ref().unwrap();
+        let right_chunk = self.chunk_buffer[*right_id].as_ref().unwrap();
+
         for sort_key_idx in &self.sort_key_idx {
-            let left_data = self.chunk_buffer[*left_id]
-                .as_ref()
-                .unwrap()
-                .array_at(*sort_key_idx)
-                .get(*left_batch_row_id);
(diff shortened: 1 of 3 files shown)


src/storage/secondary/merge_iterator.rs
```diff
@@ -48,19 +50,25 @@ impl MergeIterator {
...
+                .get(*left_batch_row_id);
+            let right_data = self.chunk_buffer[*right_id]
+                .as_ref()
+                .unwrap()
+                .array_at(*sort_key_idx)
+                .get(*right_batch_row_id);
+            // TODO: handle can-not-compare
+            res = left_data.partial_cmp(&right_data).unwrap();
+            if Ordering::Equal != res {
+                break;
```
skyzh, line 68: just return res here? so that `res` can simply be a local variable

src/storage/secondary/merge_iterator.rs
```diff
@@ -48,19 +50,25 @@ impl MergeIterator {
...
-            .array_at(self.sort_key_idx)
-            .get(*right_batch_row_id);
-
-        // TODO: handle can-not-compare
-        left_data.partial_cmp(&right_data).unwrap()
+        let mut res = Ordering::Equal;
+        for sort_key_idx in &self.sort_key_idx {
+            let left_data = self.chunk_buffer[*left_id]
+                .as_ref()
+                .unwrap()
```
skyzh, line 57: can we move this unwrap to the outer loop, so that we can do less conditional branch?

skyzh (approved):
LGTM, good work! The implementation is so precise and concise :)

src/storage/secondary/merge_iterator.rs
BaymaxHWY: Get it!

src/storage/secondary/merge_iterator.rs
BaymaxHWY: Ok

Issue #619, opened 2022-04-11
executor: refactor ValuesExecutor with DataChunkBuilder


skyzh:

JayiceZ:
Can I have a try?

skyzh:
No problem!

PR #623, opened 2022-04-13
executor: use DataChunkbuilder to refactor ValuesExecutor

JayiceZ:
close https://github.com/risinglightdb/risinglight/issues/619
Signed-off-by: Jayice <1185430411@qq.com>

### DIFF
--- a/src/executor/values.rs
+++ b/src/executor/values.rs
@@ -3,7 +3,7 @@
 use itertools::izip;
 
 use super::*;
-use crate::array::{ArrayBuilderImpl, DataChunk};
+use crate::array::{DataChunk, DataChunkBuilder};
 use crate::binder::BoundExpr;
 use crate::types::{DataType, DataTypeKind};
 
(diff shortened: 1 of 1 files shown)


src/executor/values.rs
```diff
@@ -18,44 +18,41 @@ pub struct ValuesExecutor {
...
-                    let value = expr.eval(&dummy)?;
-                    let size = match column_type.kind {
-                        Type::Varchar(size) => size,
-                        Type::Char(size) => size,
-                        _ => None,
+        type Type = DataTypeKind;
+        let mut builder = DataChunkBuilder::new(self.column_types.iter(), PROCESSING_WINDOW_SIZE);
+        let dummy = DataChunk::single(0);
+        for ref row in self.values {
+            let mut row_data: Vec<DataValue> = Vec::with_capacity(row.len());
```
xxchan, line 25: As a matter of taste, this might be _arguably_ better:
```rust
let row: Result<Vec<_>, _> = izip!(row, &self.column_types)
    .map(...)
    .collect();
if let Some(chunk) = builder.push_row(row?) {
    yield chunk;
}
```
because the `for` loop only maps the data, without other side effects.

xxchan (commented):
Others LGTM. Thanks!

src/executor/values.rs
JayiceZ: Fixed. Thank you!

xxchan (approved):
LGTM!

PR #624, opened 2022-04-13
fix: make header correct when using group by

### LINKED ISSUE #538: The header of group by is incorrect
The header is incorrect
```sql
> SELECT c0+1, c0+1+count(*) FROM t1 GROUP BY c0+1; 
+----+----------+
| c0 | ?column? |
+----+----------+
| 4  | 5        |
| 2  | 3        |
+----+----------+
in 0.004s
```

_Originally posted by @xxchan in https://github.com/risinglightdb/risinglight/issues/535#issuecomment-1055257700_

Shmiwy:
close #538
Signed-off-by: Shmiwy wyf000219@gamil.com


src/binder/expression/mod.rs
```diff
@@ -133,6 +133,26 @@ impl BoundExpr {
         visitor.visit_expr(self);
         visitor.0
     }
+
+    pub fn format_name(&self, child_schema: &Vec<ColumnDesc>) -> String {
+        println!("{:?}", self);
```
skyzh, line 138: Please don't print extra things to the screen. Also, would you please add unit tests for this function?

src/optimizer/plan_nodes/logical_aggregate.rs
```diff
@@ -69,13 +69,25 @@ impl_plan_tree_node_for_unary!(LogicalAggregate);
 impl PlanNode for LogicalAggregate {
     fn schema(&self) -> Vec<ColumnDesc> {
         let child_schema = self.child.schema();
+        let mut out_names = vec![];
         let mut input_refs = vec![];
-        self.group_keys
-            .iter()
-            .for_each(|expr| expr.resolve_input_ref(&mut input_refs));
+        self.group_keys.iter().for_each(|expr| {
```
skyzh, line 74: I don't think we need to have `input_refs` vec here. Just map group_keys to ColumnDesc, and chain it will agg_call is enough.

skyzh:
btw, you'll need to sign DCO 🤣 `git commit --amend -s && git push -f`.

src/optimizer/plan_nodes/logical_aggregate.rs
skyzh: The `resolve_input_ref` is useful in somewhere else, but not here 🤣 I forgot why we write such thing here.

Shmiwy:
> btw, you'll need to sign DCO `git commit --amend -s && git push -f`.

it seems that something terrible when I press this command(Lots of commits come up)🤣🤣🤣🤣

skyzh:
I've rebase updated for you :) You might need to git fetch && git reset --hard origin/bug_fix to make your local branch up-to-date with this remote one.

src/optimizer/plan_nodes/logical_aggregate.rs
```diff
@@ -69,13 +69,16 @@ impl_plan_tree_node_for_unary!(LogicalAggregate);
...
         self.group_keys
             .iter()
-            .for_each(|expr| expr.resolve_input_ref(&mut input_refs));
-        input_refs
-            .iter()
-            .map(|expr| child_schema[expr.index].clone())
+            .enumerate()
+            .map(|(index, expr)| {
+                ColumnDesc::new(
+                    child_schema[index].datatype().clone(),
```
skyzh, line 77: still doesn't seem correct? we should use expression's return type instead of child schema's type.

src/binder/expression/mod.rs
```diff
@@ -133,6 +133,25 @@ impl BoundExpr {
         visitor.visit_expr(self);
         visitor.0
     }
+
+    pub fn format_name(&self, child_schema: &Vec<ColumnDesc>) -> String {
```
skyzh, line 137: Please add unit tests for this function :)

skyzh (commented):
Rest LGTM, good work!

skyzh (approved):

Issue #539, opened 2022-03-01
Group by panics


xxchan:
This reports error
```sql 
> SELECT c0+2 FROM t1 GROUP BY c0+1; 
logical plan error: (Column #ColumnRefId { database_id: 0, schema_id: 0, table_id: 2, column_id: 0 } + 2) must appear in the GROUP BY clause or be used in an aggregate function
```
But this panics
```sql
> SELECT c0+2+count(*) FROM t1 GROUP BY c0+1; 
thread 'tokio-runtime-worker' panicked at 'Column #ColumnRefId { database_id: 0, schema_id: 0, table_id: 2, column_id: 0 } should not be evaluated in `eval_array`', src/executor/evaluator.rs:55:18
```

And this (similar to https://github.com/risinglightdb/risinglight/issues/501, maybe fixed by https://github.com/risinglightdb/risinglight/discussions/503)
```sql
> select c0*1 from t1 group by c0*1
+----+
| c0 |
+----+
| 3  |
| 1  |
+----+
in 0.004s
> select c0*0 from t1 group by c0*0
thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 0 but the index is 0', src/optimizer/plan_nodes/logical_projection.rs:66:25
```

_Originally posted by @xxchan in https://github.com/risinglightdb/risinglight/pull/535#issuecomment-1055257700_

skyzh:
There will be more and more edge cases after we support subquery... may need a better design ahead of implementation for new PRs.

skyzh:
I guess now it won't panic, but will only report an error. Let's get this issue closed.

PR #630, opened 2022-04-17
refactor(planner): use AggInputRefResolver to handle group by expr

Fedomn:
fix part of https://github.com/risinglightdb/risinglight/issues/539

the original group by expression support (https://github.com/risinglightdb/risinglight/pull/535) seems not very clear for resolving `InputRef`, so I refactor the main logic to `InputRefResolver`. And fix panic for:
```sql 
select v2 + 1 + count(*) from t group by v2 + 1;
```

but it still leaves some issues:

- panic `select v2 + 2 + count(*) from t group by v2 + 1;`

### DIFF
--- a/tests/sql/group_by.slt
+++ b/tests/sql/group_by.slt
@@ -21,26 +21,35 @@ select v2 + 1, sum(v1) from t group by v2 + 1
 2 3
 3 7
 
-query II rowsort
+query III rowsort
 select sum(v1), v2 + 1 as a, count(*) from t group by a
 ----
 3 2 2
 7 3 2
 5 4 1
 
-query II rowsort
+query III rowsort
 select v2, v2 + 1, sum(v1) from t group by v2 + 1, v2
 ----
 1 2 3
 3 4 5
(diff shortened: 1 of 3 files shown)


tests/sql/group_by.slt
```diff
@@ -21,26 +21,35 @@ select v2 + 1, sum(v1) from t group by v2 + 1
...
-query II rowsort
+query III rowsort
 select sum(v1), v2 + 1 as a, count(*) from t group by a
 ----
 3 2 2
 7 3 2
 5 4 1
 
-query II rowsort
+query III rowsort
```
skyzh, line 31: We need to verify if provided column number is the same as user input? cc @wangrunji0408

skyzh (approved):
That's cool, thanks!

tests/sql/group_by.slt
wangrunji0408: I believe it should verify. But now the database returns a string, which makes it hard to split columns (

Issue #625, opened 2022-04-14
catalog: support composite primary key


skyzh:
Currently, our catalog cannot handle the case for composite primary key. For example,

```
create table t1 (x int not null, y int not null, primary key(x, y));
create table t2 (x int not null, y int not null, primary key(y, x));
```

These two tables will have exactly the same catalog entry, despite that the first one should have (x, y) as composite sort key, and the second one have (y, x). The order is different!

Therefore, we will need to refactor the catalog to correctly record such information. For `TableCatalog`, we should have a field `primary_key_indices`, and remove the `is_primary` field from `ColumnDesc`.

This task might involve some refactors and some regressions in functionalities. Need plan first.

skyzh:
... may also be done together with https://github.com/risinglightdb/risinglight/issues/590

skyzh:
assigned to @Shmiwy

Shmiwy:
It seems that we don't support primary key? when I type this sql
```sql
create table t (a int not null, b int not null, c int, d int, primary key(a, b));
```
the physicalCreateTable says that neither a nor b is primary key🤣
<img width="1072" alt="image" src="https://user-images.githubusercontent.com/65114384/163789788-ae1f2b6c-e83f-464e-be48-3f2c477fde91.png">

skyzh:
Then our binder doesn't support such syntax now 🤣 Please also help refactor the binder to support primary key (a, b) syntax.

https://github.com/risinglightdb/risinglight/blob/a5644b13621063b4f38f45155defaf8157498739/src/binder/statement/create_table.rs#L20-L22

For example, the first statement produces AST like:

```
        CreateTable {
            or_replace: false,
            temporary: false,
            external: false,
            if_not_exists: false,
            name: ObjectName([Ident {
                value: "t1",
                quote_style: None,
            }]),
            columns: [
                ColumnDef {
                    name: Ident {
                        value: "x",
                        quote_style: None,
                    },
                    data_type: Int(None),
                    collation: None,
                    options: [ColumnOptionDef {
                        name: None,
                        option: NotNull,
                    }],
                },
                ColumnDef {
                    name: Ident {
                        value: "y",
                        quote_style: None,
                    },
                    data_type: Int(None),
                    collation: None,
                    options: [ColumnOptionDef {
                        name: None,
                        option: NotNull,
                    }],
                },
            ],
            constraints: [Unique {
                name: None,
                columns: [
                    Ident {
                        value: "x",
[...]

PR #631, opened 2022-04-19
feat(binder): support `primary key (a, b)` syntax

Shmiwy:

#625 1st editon

support primary key (a, b) syntax by adding a field called `primary_key_ids` in `TableCatalog` without removng `is_primary field` from `ColumnDesc`. I will remove `is_primary` and refactor code when I want to extend the case to multiple sort keys.

### DIFF
--- a/src/catalog/table.rs
+++ b/src/catalog/table.rs
@@ -21,6 +21,7 @@ struct Inner {
     #[allow(dead_code)]
     is_materialized_view: bool,
     next_column_id: ColumnId,
+    primary_key_ids: Vec<ColumnId>,
 }
 
 impl TableCatalog {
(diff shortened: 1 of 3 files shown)


src/catalog/table.rs
```diff
@@ -38,14 +39,31 @@ impl TableCatalog {
...
         };
+        let mut pk_ids = vec![];
         for col_catalog in columns {
+            if col_catalog.is_primary() {
+                pk_ids.push(col_catalog.id());
+            }
             table_catalog.add_column(col_catalog).unwrap();
         }
+
+        table_catalog.set_primary_key_ids(&pk_ids);
```
skyzh, line 53: It seems that this logic still cannot tell the different between `primary key (a, b)` and `primary key (b, a)`? May fix this in later PRs.

skyzh (approved):
LGTM, good work!

Shmiwy:
@skyzh 

primary_key_ids is vector， maybe we can tell the difference by index？
https://github.com/risinglightdb/risinglight/blob/d28389a62069cee31882dcae4a4e10383e4518f2/src/catalog/table.rs#L15-L25

skyzh:
But when generating `primary_key_ids` in `TableCatalog::new`, we simply iterate from the first column to the end. It seems that `Unique` constraint is not retained?

src/catalog/table.rs
Shmiwy: `primary_key_ids` is a vector， maybe we can tell the difference by index？

Shmiwy:
oh, I forget this...

skyzh:
> oh, I forget this...

May also add a test case for this in the future :)

PR #633, opened 2022-04-19
feat: support multi ordered primary key

Shmiwy:

try to implementation #625

at this moment, I just append `ordered_pk_ids` to `tableCatalog` withoud using  `is_primary` filed in `ColumnDesc`. It will make removing  `is_primary` in the future easier.

### DIFF
--- a/src/binder/statement/create_table.rs
+++ b/src/binder/statement/create_table.rs
@@ -91,8 +91,8 @@ impl Binder {
         }
     }
 
-    // get primary keys' id in declared order。
-    // we use index in columns vector as column id
+    /// get primary keys' id in declared order。
+    /// we use index in columns vector as column id
     fn ordered_pks_from_columns(columns: &[ColumnDef]) -> Vec<ColumnId> {
         let mut ordered_pks = Vec::new();
 
@@ -111,9 +111,9 @@ impl Binder {
         ordered_pks
     }
 
-    // We have used `pks_name_from_constraints` to get the primary keys' name sorted by declaration
-    // order in "primary key(c1, c2..)" syntax. Now we transfer the name to id to get the sorted
-    // ID
+    /// We have used `pks_name_from_constraints` to get the primary keys' name sorted by declaration
+    /// order in "primary key(c1, c2..)" syntax. Now we transfer the name to id to get the sorted
+    /// ID
     fn ordered_pks_from_constraint(pks_name: &[String], columns: &[ColumnDef]) -> Vec<ColumnId> {
         let mut ordered_pks = vec![0; pks_name.len()];
         let mut pos_in_ordered_pk = HashMap::new(); // used to get pos from column name
@@ -131,7 +131,7 @@ impl Binder {
         });
         ordered_pks
     }
-    // get the primary keys' name sorted by declaration order in "primary key(c1, c2..)" syntax.
+    /// get the primary keys' name sorted by declaration order in "primary key(c1, c2..)" syntax.
     fn pks_name_from_constraints(constraints: &[TableConstraint]) -> Vec<String> {
         let mut pks_name_from_constraints = vec![];
 
--- a/src/binder/statement/create_table.rs
+++ b/src/binder/statement/create_table.rs
@@ -45,40 +45,45 @@ impl Binder {
                     }
                 }
 
-                // column name -> index in `ordered_pk_ids` vector
-                let ordered_pk_pos = Binder::get_extra_pks(constraints);
-                // whether we have primary key declared by "primary key(c1, c2..) syntax"
-                let flag = !ordered_pk_pos.is_empty();
-                // ordered primary ids declared by "primary key(c1, c2..) syntax"
-                let mut ordered_pk_ids = vec![0; ordered_pk_pos.len()];
-                let mut col_catalogs = vec![];
+                let mut ordered_pk_ids = Binder::ordered_pks_from_columns(columns);
+                let has_pk_from_column = !ordered_pk_ids.is_empty();
 
-                for (idx, colum_def) in columns.iter().enumerate() {
-                    let mut is_primary = false;
-                    let mut col = ColumnCatalog::from(colum_def, &mut is_primary);
-                    if flag && is_primary {
-                        // we can't support sql query like
-                        // "create table (a int primary key, b int not null, primary key(b));"
-                        // declaring pk both in column's option and 'primary key(c1..)' syntax"
-                        return Err(BindError::NotSupportedTSQL);
-                    } else if is_primary {
-                        // primary key is declared only in column's options
-                        ordered_pk_ids.push(idx as ColumnId);
-                    } else if ordered_pk_pos.contains_key(col.name()) {
-                        // primary key is declared only by "primary key(c1, c2) syntax"
-                        let pos = *(ordered_pk_pos.get(col.name()).unwrap());
-                        ordered_pk_ids[pos] = idx as ColumnId;
-                        col.set_primary(true); // TODO: remove this line in the future
-                    }
-                    col.set_id(idx as ColumnId);
-                    col_catalogs.push(col);
+                if ordered_pk_ids.len() > 1 {
+                    // multi primary key should be declared by "primary key(c1, c2...)" syntax
+                    return Err(BindError::NotSupportedTSQL);
+                }
+
+                let pks_name_from_constraints = Binder::pks_name_from_constraints(constraints);
+                if has_pk_from_column && !pks_name_from_constraints.is_empty() {
+                    // can't get primary key both from "primary key(c1, c2...)" syntax and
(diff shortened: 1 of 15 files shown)


src/binder/statement/create_table.rs
```diff
@@ -45,57 +45,73 @@ impl Binder {
...
+        for constraint in constraints {
+            match constraint {
+                TableConstraint::Unique {
+                    is_primary,
+                    columns,
+                    ..
+                } if *is_primary => columns.iter().enumerate().for_each(|(index, indent)| {
+                    ordered_pk_pos.insert(indent.value.clone(), index);
+                }),
+                _ => todo!(),
```
skyzh, line 100: Better to return an error, so that the program won't panic.

Also, will there be multiple `unique` constraints in parser AST? We should only allow single unique appearing in a statement, and return error otherwise.

skyzh: Also I think the signature of this function is somehow confusing... The `HashMap<String, usize>` usize part seems like position of column in unique constraint. But generally, developers want a column id at this stage. For example,

```
create table (v1 int, v2 int, primary key(v2));
```

... where currently we will generate v2 -> 0. But what we need is something like `pk: vec![column id #1]`.

I think we can simply return Vec<Ident> for this function, and let the other part of the code to handle the mapping between ident and column id.

skyzh: Also I think the signature of this function is confusing... The `HashMap<String, usize>` usize part seems like position of column in unique constraint. But generally, developers want a column id at this stage. For example,

```
create table (v1 int, v2 int, primary key(v2));
```

... where currently we will generate v2 -> 0. But what we need is something like `pk: vec![column id #1]`.

I think we can simply return Vec<Ident> for this function, and let the other part of the code to handle the mapping between ident and column id.

src/binder/statement/create_table.rs
```diff
@@ -124,10 +140,14 @@ mod tests {
         let catalog = Arc::new(RootCatalog::new());
         let mut binder = Binder::new(catalog.clone());
         let sql = "
-            create table t1 (v1 int not null, v2 int); 
+            create table t1 (v1 int not null, v2 int);
```
skyzh, line 143: We can have planner test in the future 🤣 I guess writing this test case must be painful.

FYI, https://github.com/singularity-data/risingwave/blob/main/src/frontend/test_runner/tests/testdata/tpch.yaml

skyzh, line 143: We can have planner test in the future 🤣 I guess writing this test case must be painful.

FYI, https://github.com/singularity-data/risingwave/blob/main/src/frontend/test_runner/tests/testdata/tpch.yaml

src/binder/statement/create_table.rs
```diff
@@ -45,57 +45,73 @@ impl Binder {
...
-                            ..
-                        } if *is_primary => columns.iter().for_each(|indent| {
-                            extra_pk_name.insert(indent.value.clone());
-                        }),
-
-                        _ => todo!(),
+                // column name -> index in `ordered_pk_ids` vector
+                let ordered_pk_pos = Binder::get_extra_pks(constraints);
+                // whether we have primary key declared by "primary key(c1, c2..) syntax"
+                let flag = !ordered_pk_pos.is_empty();
```
skyzh, line 51: flag -> has_pk_constraint

skyzh, line 51: flag -> has_pk_constraint

src/binder/statement/create_table.rs
```diff
@@ -45,57 +45,73 @@ impl Binder {
...
+            }
+        }
+        ordered_pk_pos
+    }
 }
 
-impl From<&ColumnDef> for ColumnCatalog {
-    fn from(cdef: &ColumnDef) -> Self {
+impl ColumnCatalog {
+    fn from(cdef: &ColumnDef, is_primary_: &mut bool) -> Self {
```
skyzh, line 108: Generally in the Rust programming language, `from` will only take one parameter. I'd suggest not doing two things at once in `from`. Just declare it as

`from(cdef: &ColumnDef) -> Self`

and access `is_primary` field for this column later.

skyzh, line 108: Generally in the Rust programming language, `from` will only take one parameter. I'd suggest not doing two things at once in `from`. Just declare it as

`from(cdef: &ColumnDef) -> Self`

and access `is_primary` field for this column later.

skyzh: ... and I guess we can simply ignore `ColumnOption::Unique` here, and add a helper function to determine whether a `ColumnDef` is a primary key?

skyzh: ... and I guess we can simply ignore `ColumnOption::Unique` here, and add a helper function to determine whether a `ColumnDef` is a primary key?

src/binder/statement/create_table.rs
```diff
@@ -45,57 +45,73 @@ impl Binder {
...
+                    let mut is_primary = false;
+                    let mut col = ColumnCatalog::from(colum_def, &mut is_primary);
+                    if flag && is_primary {
+                        // we can't support sql query like
+                        // "create table (a int primary key, b int not null, primary key(b));"
+                        // declaring pk both in column's option and 'primary key(c1..)' syntax"
+                        return Err(BindError::NotSupportedTSQL);
+                    } else if is_primary {
+                        // primary key is declared only in column's options
+                        ordered_pk_ids.push(idx as ColumnId);
```
skyzh, line 66: So the default behavior of `create table (a int primary key, b int primary key)` will be a composite primary key of `(a, b)`? From my perspective, it's better to only allow one single pk in this case, so that users must specify whether it's `(a, b)` or `(b, a)` for composite pk?

skyzh, line 66: So the default behavior of `create table (a int primary key, b int primary key)` will be a composite primary key of `(a, b)`? From my perspective, it's better to only allow one single pk in this case, so that users must specify whether it's `(a, b)` or `(b, a)` for composite pk?

src/storage/mod.rs
```diff
@@ -75,6 +75,7 @@ pub trait Storage: Sync + Send + 'static {
         schema_id: SchemaId,
         table_name: &'a str,
         column_descs: &'a [ColumnCatalog],
+        ordered_pk_ids: Vec<ColumnId>,
```
skyzh, line 78: To ensure consistency with other fields, better to also pass &[ColumnId] here 🤣

skyzh, line 78: To ensure consistency with other fields, better to also pass &[ColumnId] here 🤣

src/storage/secondary/manifest.rs
```diff
@@ -164,7 +166,12 @@ impl SecondaryStorage {
             return Err(TracedStorageError::duplicated("table", table_name));
         }
         let table_id = schema
-            .add_table(table_name.clone(), column_descs.to_vec(), false)
+            .add_table(
+                table_name.clone(),
+                column_descs.to_vec(),
+                false,
+                ordered_pk_ids,
```
skyzh, line 173: `ordered_pk_ids` also needs to be persisted to disk in manifest.rs `CreateTableEntry`. We can do this in later PRs.

skyzh, line 173: `ordered_pk_ids` also needs to be persisted to disk in manifest.rs `CreateTableEntry`. We can do this in later PRs.

skyzh (commented):
Generally LGTM, good work!

For the `bind_create_table` part, I suggest to reorganize the code, so that it can be much clearer to readers.

* Firstly, get the pk constraints ident vec from `get_extra_pks`, like `vec!["a"]`
* Then, replace vec of ident to vec of column id, e.g., `vec![ColumnId(1)]`. May use hashmap in the process to accelerate lookups. We call this `ordered_pks_from_constraint`.
* After that, iterate all columns, and generate a new set of `ordered_pks_from_columns` from `is_primary` property.
* If both `ordered_pks_from_constraint` and `ordered_pks_from_columns` are not empty, return error. If `ordered_pks_from_columns` has more than 1 elements, return error (if you want).
* Then we can get the final `ordered_pks` from either `ordered_pks_from_constraint` or `ordered_pks_from_columns`.

src/binder/statement/create_table.rs
```diff
@@ -45,73 +45,121 @@ impl Binder {
...
+                };
+                if is_primary_ {
+                    ordered_pks.push(index as ColumnId);
+                }
+            }
+        }
+        ordered_pks
+    }
+
+    // We have used `pks_name_from_constraints` to get the primary keys' name sorted by declaration
```
skyzh, line 114: Use `///` for comments on functions, so that Rust will recognize it as docstring, and you can view it in `make docs`.

skyzh (approved):
Rest LGTM, good work!

PR #637, opened 2022-04-23
refactor(executor): refactor CopyFromFile using DataChunkBuilder

### LINKED ISSUE #618: executor: refactor CopyFromFile using DataChunkBuilder

noneback:
refactor CopyFromFile using DataChunkBuilder

close #618

### DIFF
--- a/src/array/data_chunk_builder.rs
+++ b/src/array/data_chunk_builder.rs
@@ -3,7 +3,7 @@ use std::iter::IntoIterator;
 use itertools::Itertools;
 
 use super::{ArrayBuilderImpl, DataChunk};
-use crate::types::{DataType, DataValue};
+use crate::types::{ConvertError, DataType, DataValue};
 
 /// A helper struct to build a [`DataChunk`].
 ///
@@ -54,6 +54,31 @@ impl DataChunkBuilder {
             self.take()
         } else {
             None
+        }
+    }
+
+    /// Push a row of str in the Iterator.
+    ///
+    /// The row is accepted as an iterator of [`&str`], and it's required that the size of row
+    /// should be the same as the number of columns.
+    ///
+    /// It will convert [`&str`] into specified type, and a [`ConvertError`] will be returned if
+    /// converting fails.
+    ///
+    /// A [`DataChunk`] will be returned while converting correctly and `size == capacity`.
+    pub fn push_str_row<'a>(
+        &mut self,
+        row: impl IntoIterator<Item = &'a str>,
+    ) -> Result<Option<DataChunk>, ConvertError> {
+        for (builder, r) in self.array_builders.iter_mut().zip_eq(row) {
+            builder.push_str(r)?
+        }
+
+        self.size += 1;
+        if self.size == self.capacity {
+            Ok(self.take())
+        } else {
+            Ok(None)
         }
     }
 
--- a/src/executor/copy_from_file.rs
+++ b/src/executor/copy_from_file.rs
@@ -104,13 +104,11 @@ impl CopyFromFileExecutor {
                         Ok(v)
                     })
                     .collect();
-
-            // update progress bar
             size_count += record.as_slice().as_bytes().len();
-            bar.set_position(size_count as u64);
 
             // push a raw str row and send it if necessary
             if let Some(chunk) = chunk_builder.push_str_row(str_row_data?)? {
+                bar.set_position(size_count as u64);
                 if token.is_cancelled() || tx.blocking_send(chunk).is_err() {
                     return Err(ExecutorError::Abort);
                 }
--- a/src/executor/copy_from_file.rs
+++ b/src/executor/copy_from_file.rs
@@ -77,7 +77,7 @@ impl CopyFromFileExecutor {
 
         let column_count = self.plan.logical().column_types().len();
 
-        // create chunk builders
+        // create chunk builder
         let mut chunk_builder =
             DataChunkBuilder::new(self.plan.logical().column_types(), PROCESSING_WINDOW_SIZE);
         let mut size_count = 0;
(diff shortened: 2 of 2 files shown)


src/executor/copy_from_file.rs
```diff
@@ -78,14 +79,9 @@ impl CopyFromFileExecutor {
...
-                .column_types()
-                .iter()
-                .map(|ty| ArrayBuilderImpl::with_capacity(PROCESSING_WINDOW_SIZE, ty))
-                .collect_vec();
+            // create chunk builders
+            let mut chunk_builder =
+                DataChunkBuilder::new(self.plan.logical().column_types(), PROCESSING_WINDOW_SIZE);
 
             // read records and push to array builder
             for _ in 0..PROCESSING_WINDOW_SIZE {
```
xxchan, line 87: I think `chunk_builder` can be created outside `while !finished` loop, and `for _ in 0..PROCESSING_WINDOW_SIZE` loop is not needed any more?

skyzh: +1

src/array/data_chunk_builder.rs
```diff
@@ -57,6 +57,31 @@ impl DataChunkBuilder {
         }
     }
 
+    /// Push a row of str in the Iterator.
+    ///
+    /// The row is accepted as an iterator of &str, and it's required that the size of row
+    /// should be the same as the number of columns.
+    ///
+    /// A [`DataChunk`] will be returned while `size == capacity`, and it should always be handled
+    /// correctly.
```
xxchan, line 66:
```suggestion
    /// Push a row of `str` in the Iterator.
    ///
    /// The row is accepted as an iterator of `&str`, and it's required that the size of row
    /// should be the same as the number of columns.
    ///
    /// A [`DataChunk`] will be returned while `size == capacity`, and it should always be handled
    /// correctly.
    #[must_use]
```

src/array/data_chunk_builder.rs
```diff
@@ -57,6 +57,31 @@ impl DataChunkBuilder {
...
+    pub fn push_str_row<'a>(
+        &mut self,
+        row: impl IntoIterator<Item = &'a str>,
+    ) -> Result<Option<DataChunk>, ConvertError> {
+        for (builder, r) in self.array_builders.iter_mut().zip_eq(row) {
+            if let Some(e) = builder.push_str(r).err() {
+                return Err(e);
+            }
```
xxchan, line 74: A style suggestion:
```suggestion
            builder.push_str(r)?;
```

src/executor/copy_from_file.rs
noneback: Ok, I will try to fix it. I am not so familiar with rust. 
Sorry for that.

src/executor/copy_from_file.rs
```diff
@@ -104,27 +100,33 @@ impl CopyFromFileExecutor {
...
 
             // send data chunk
-            let chunk: DataChunk = array_builders.into_iter().collect();
-
             #[allow(clippy::collapsible_if)]
-            if chunk.cardinality() > 0 {
-                if token.is_cancelled() || tx.blocking_send(chunk).is_err() {
-                    return Err(ExecutorError::Abort);
+            if let Some(chunk) = { chunk_builder }.take() {
+                if chunk.cardinality() > 0 {
```
xxchan, line 126: From the doc of `DataChunkBuilder::take`: If there are no remaining rows, `None` will be returned.

So I guess `if chunk.cardinality() > 0 {` can be removed. (And maybe also `#[allow(clippy::collapsible_if)]`?)

src/executor/copy_from_file.rs
xxchan: No need to be sorry. Just try it! 😄 

You may also refer to other refactors e.g. https://github.com/risinglightdb/risinglight/pull/623/files to see how the pattern changes

noneback: Thx for your patience! It's of great help:)

xxchan (commented):
Btw the tests failed. I haven't taken a look yet. You may try to fix it

src/array/data_chunk_builder.rs
noneback: `Result` has already been marked as `#[must_use]`, so I guess there is no need to add one.

src/executor/copy_from_file.rs
```diff
@@ -75,59 +76,58 @@ impl CopyFromFileExecutor {
...
+                        }
+                        Ok(v)
+                    })
+                    .collect();
+
+            // push a raw str row
+            let chunk = match chunk_builder.push_str_row(str_row_data?) {
+                Ok(chunk) => chunk,
+                Err(e) => return Err(ExecutorError::Convert(e)),
+            };
```
xxchan, line 112: Can be simplified and merged below
```suggestion
```

src/executor/copy_from_file.rs
```diff
@@ -75,59 +76,58 @@ impl CopyFromFileExecutor {
...
 
-            #[allow(clippy::collapsible_if)]
-            if chunk.cardinality() > 0 {
+            // update progress bar
+            size_count += record.as_slice().as_bytes().len();
+            bar.set_position(size_count as u64);
+
+            // send chunk if necessary
+            if let Some(chunk) = chunk {
```
xxchan, line 119:
```suggestion
            if let Some(chunk) = chunk_builder.push_str_row(str_row_data?)?; {
```

xxchan (approved):
Others LGTM!

src/executor/copy_from_file.rs
```diff
@@ -75,59 +76,52 @@ impl CopyFromFileExecutor {
...
+                        }
+                        Ok(v)
+                    })
+                    .collect();
 
-            #[allow(clippy::collapsible_if)]
-            if chunk.cardinality() > 0 {
+            // update progress bar
+            size_count += record.as_slice().as_bytes().len();
+            bar.set_position(size_count as u64);
```
skyzh, line 110: update bar per-record will have significant performance impact. Consider only updating it when a chunk is produced.

skyzh (commented):
Rest LGTM, good work!

skyzh (approved):
Thanks!

PR #642, opened 2022-04-27
fix(planner): enhance group by illegal column validation

Fedomn:
fix part of https://github.com/risinglightdb/risinglight/issues/539

including two enhanced validation for `groupby` illegal columns:
- `select v2 + 2 + count(*) from t group by v2 + 1;`
- `select v2 + count(*) from t group by v2 order by v1;`


skyzh (approved):
That's cool, thanks!

PR #644, opened 2022-05-04
feat(storage): support range-filter scan by sort key

### LINKED ISSUE #589: storage: support range-filter scan by sort key

Shmiwy:
close #589
for now only support range-filter scan by first column and this column should by primary key

### DIFF
--- a/src/storage/secondary/rowset/disk_rowset.rs
+++ b/src/storage/secondary/rowset/disk_rowset.rs
@@ -120,8 +120,9 @@ impl DiskRowset {
         dvs: Vec<Arc<DeleteVector>>,
         seek_pos: ColumnSeekPosition,
         expr: Option<BoundExpr>,
+        end_sort_key: Option<&[u8]>,
     ) -> StorageResult<RowSetIterator> {
-        RowSetIterator::new(self.clone(), column_refs, dvs, seek_pos, expr).await
+        RowSetIterator::new(self.clone(), column_refs, dvs, seek_pos, expr, end_sort_key).await
     }
 
     pub fn on_disk_size(&self) -> u64 {
@@ -130,6 +131,32 @@ impl DiskRowset {
             .map(|x| x.on_disk_size())
             .sum1()
             .unwrap_or(0)
+    }
+
+    /// get the start row id to begin with for later table scanning.
+    /// if `begin_sort_key` is none, we return `ColumnSeekPosition::RowId(0)` to indicate scanning
+    /// from the beginning, otherwise we scan the rowset's first column indexes, find the first
+    /// block who contains data greater than or equal to `begin_sort_key` and return the row id of
+    /// the block's first key. Currently, only the first column of the rowset can be used to get
+    /// the start row id and this column should be primary key.
+    pub async fn start_row_id(&self, begin_sort_key: Option<&[u8]>) -> ColumnSeekPosition {
+        if begin_sort_key.is_none() {
+            return ColumnSeekPosition::RowId(0);
+        }
+        let begin_sort_key = begin_sort_key.unwrap();
+        let column = self.column(0);
+        let column_index = column.index();
+
+        let mut row_id: u32 = 0;
+        for index in column_index.indexes() {
+            if index.first_key.as_slice() > begin_sort_key {
+                // cur block's first key is greater then `begin_sort_key`
+                // which indicate that we should scan from the last block
+                break;
+            }
+            row_id = index.first_rowid;
+        }
+        ColumnSeekPosition::RowId(row_id)
     }
 }
 
--- a/src/storage/secondary/rowset/disk_rowset.rs
+++ b/src/storage/secondary/rowset/disk_rowset.rs
@@ -1,5 +1,6 @@
 // Copyright 2022 RisingLight Project Authors. Licensed under Apache-2.0.
 
+use std::borrow::Borrow;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 
@@ -14,8 +15,10 @@ use super::{path_of_data_column, path_of_index_column, RowSetIterator};
 use crate::binder::BoundExpr;
 use crate::catalog::ColumnCatalog;
 use crate::storage::secondary::column::ColumnReadableFile;
+use crate::storage::secondary::encode::PrimitiveFixedWidthEncode;
 use crate::storage::secondary::DeleteVector;
 use crate::storage::{StorageColumnRef, StorageResult};
+use crate::types::DataValue;
 
 /// Represents a column in Secondary.
 ///
@@ -120,7 +123,7 @@ impl DiskRowset {
         dvs: Vec<Arc<DeleteVector>>,
         seek_pos: ColumnSeekPosition,
         expr: Option<BoundExpr>,
-        end_sort_key: Option<&[u8]>,
+        end_sort_key: &[DataValue],
     ) -> StorageResult<RowSetIterator> {
         RowSetIterator::new(self.clone(), column_refs, dvs, seek_pos, expr, end_sort_key).await
     }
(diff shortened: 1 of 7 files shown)


skyzh (commented):
Would you please add some unit tests for `RowSetIterator` for range filters before starting review? 🤪

Shmiwy:
> Would you please add some unit tests for `RowSetIterator` for range filters before starting review? 🤪

oh， I forgot again....🥲

src/storage/secondary/rowset/disk_rowset.rs
```diff
@@ -120,8 +123,9 @@ impl DiskRowset {
         dvs: Vec<Arc<DeleteVector>>,
         seek_pos: ColumnSeekPosition,
         expr: Option<BoundExpr>,
+        end_sort_key: &[DataValue],
     ) -> StorageResult<RowSetIterator> {
-        RowSetIterator::new(self.clone(), column_refs, dvs, seek_pos, expr).await
+        RowSetIterator::new(self.clone(), column_refs, dvs, seek_pos, expr, end_sort_key).await
```
skyzh, line 128: Why we only pass `end_sort_key` here? I guess this function should have both `begin_sort_key` and `end_sort_key` as parameter?

src/storage/secondary/rowset/disk_rowset.rs
```diff
@@ -131,10 +135,52 @@ impl DiskRowset {
...
+                        break;
+                    }
+                    pre_block_first_key = index.first_rowid;
+                }
+                pre_block_first_key
+            }
+            // Todo: support ohter type
+            _ => panic!("for now support range-filter scan by sort key type of int32"),
+        };
+        ColumnSeekPosition::RowId(start_row_id)
```
skyzh, line 176: This positioning is not accurate. If we have 2 blocks with start of:

```
block 1: a
block 2: d
```

Now we want to seek to `b`. This function will return row_id = 0. However, the first row from block 1 cannot be `b` -- it's `a` instead. Therefore, `RowSetIterator` will still need `begin_sort_key` as parameter, and it will need to filter those keys below start sort key.

src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -265,14 +275,46 @@ impl RowSetIterator {
...
             }
+            // For now, we only support range-filter scan by first column.
+            if !self.end_keys.is_empty() && id == 0 {
+                let end_key = &self.end_keys[0];
+                let array = arrays[0].as_ref().unwrap();
+                let len = array.len();
+
+                // Todo: only suppor range-filter scan by sort key type of int32, support other type
+                // later.
+                if end_key - &array.get(len - 1) < DataValue::Int32(0) {
```
skyzh, line 286: Directly compare by `end_key < &array.get(len - 1)`?

skyzh (commented):
The logic generally correct, but there are some small issues. Comments below.

Also one idea come in my mind: maybe we can make range filter part of the filter condition, so that we don't need to compute which rows should be removed by ourselves, and re-use the current expression framework?

We can doc our RowSetIterator and Scan interface like: begin / end sort key must be included in filter condition as `10 < x and x < 15`, so as to make range filter scan work correctly.

Now we only need to do two additional things compared with the previous code:

1. seek to the block (no need to be accurate, as filter condition will help us filter rows out-of-range)
2. pre-compute the block that contains `end_sort_key` instead of checking the last item in the data chunk. That's because filter condition will help us filter rows out-of-range, and it might incur a lot I/Os before a row can be produced. So if we stop at the block instead of checking every data chunk, things could be easier.

For example, let take this SQL as example:

```
select * from table where x > 10 and x < 15;
```

The RowSetIterator will be created as:

begin_sort_key: [10], end_sort_key: [15], filter: 10 < x and x < 15

Before creating RowSetIterator, we will position the first block with `10` occurrence (namely block A), and the last block with `15` occurrence (namely block B). At the beginning, the RowSetIterator will be positioned at the first row of Block A. When the RowSetIterator goes beyond block B (we can check this by current row id > last row of block B), we return the end of iterator.

Therefore, things could be a lot easier. Also, we can easily support `>` and `>=` conditions.

src/storage/secondary/rowset/disk_rowset.rs
Shmiwy: because i put the start row id in seek_pos, and will finish scanning when meet `end_sort_key`

src/storage/secondary/rowset/disk_rowset.rs
Shmiwy: oh .the line 166 should be >= instead of >

skyzh: I mean the begin row might not be the same as sort key? In this case, the RowSetIterator might produce more items than expected. You may take a look at the above comment, and put range filters into the filter condition when unit testing.

src/storage/secondary/rowset/rowset_iterator.rs
Shmiwy: Because I found that we implemented  addition, subtraction, multiplication and division between `DataValue`（like int32， int64). so i try reuse the logic to avoid many   forced conversions

src/storage/secondary/rowset/disk_rowset.rs
Shmiwy: oh yes. i forget this

src/storage/secondary/rowset/rowset_iterator.rs
Shmiwy: This thing is a little weird. I'll think of something else.😂

Shmiwy:
just fix some bug in this commit, I will make range filter part of the filter condition in the later pr

src/storage/secondary/rowset/disk_rowset.rs
```diff
@@ -131,10 +145,52 @@ impl DiskRowset {
...
+    /// Get the start row id to begin with for later table scanning.
+    /// If `begin_keys` is empty, we return `ColumnSeekPosition::RowId(0)` to indicate scanning
+    /// from the beginning, otherwise we scan the rowsets' first column indexes, find the first
+    /// block who contains data greater than or equal to `begin_key` and return the row id of
+    /// the block's first key. Currently, only the first column of the rowsets can be used to get
+    /// the start row id and this column should be primary key.
+    /// If `begin_key` is greater than all blocks' `first_key`, we return the `first_key` of the
+    /// last block.
+    /// Todo: support multi sort-keys range filter
+    pub async fn start_rowid(&self, begin_keys: &[DataValue]) -> ColumnSeekPosition {
```
skyzh, line 158: So we are assuming that the first column of the RowSet is pk, and pk is non-nullable. Should enforce this constraint in binder in later PRs.

src/storage/secondary/rowset/disk_rowset.rs
```diff
@@ -131,10 +145,52 @@ impl DiskRowset {
...
+                let mut pre_block_first_key = 0;
+                for index in column_index.indexes() {
+                    let mut first_key: &[u8] = &index.first_key;
+                    let first_val: i32 = PrimitiveFixedWidthEncode::decode(&mut first_key);
+
+                    if first_val > begin_val {
+                        break;
+                    }
+                    pre_block_first_key = index.first_rowid;
+                }
```
skyzh, line 180: One optimization can be done in the future PRs:

* Use `partition_point` function to do binary search, which could be a lot faster.

src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -95,13 +101,22 @@ impl RowSetIterator {
             dvs,
             column_iterators,
             filter_expr,
+            start_keys: start_keys.to_vec(),
```
skyzh, line 104: I think we can simply pass `begin_row_id` and `end_row_id` into RowSetIterator? So that we don't need complex logic to compare chunk against sort key in `next_batch`. `begin_row_id` and `end_row_id` include the range to scan, and we use filter scan to filter data. For example, the outer function determines that block 2 - 3 contains data for user specified sort key. So we can pass block 2's first row_id as begin_row_id, and block 4's first row_id - 1 as end_row_id.

src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -450,4 +543,293 @@ mod tests {
...
+                .iter(
+                    vec![
+                        StorageColumnRef::Idx(0),
+                        StorageColumnRef::Idx(1),
+                        StorageColumnRef::Idx(2),
+                    ]
+                    .into(),
+                    vec![],
+                    ColumnSeekPosition::RowId(168),
+                    None,
```
skyzh, line 564: Filter condition should include the range info. Here we can construct an expression manually -- `180 < InputRef(0) < 195`.

skyzh (commented):
Rest LGTM!

skyzh:
> just fix some bug in this commit, I will make range filter part of the filter condition in the later pr

Okay, let me re-do some reviews. If you make range filter part of the filter condition, there are a lot of code that can be removed later :)

src/storage/secondary/rowset/rowset_iterator.rs
skyzh: May fix in later PRs.

src/storage/secondary/rowset/rowset_iterator.rs
skyzh: May fix in later PRs.

src/storage/secondary/rowset/rowset_iterator.rs
```diff
@@ -265,14 +280,54 @@ impl RowSetIterator {
...
+            // For now, we only support range-filter scan by first column.
+            if id == 0 {
+                if !self.start_keys.is_empty() && !self.meet_start_key_before {
+                    // find the first row in range to begin with
+                    self.meet_start_key_before = true;
+                    let array = arrays[0].as_ref().unwrap();
+                    let len = array.len();
+                    let start_key = &self.start_keys[0];
+                    let start_row_id =
+                        (0..len).position(|idx| start_key - &array.get(idx) <= DataValue::Int32(0));
```
skyzh, line 292: Can do a binary search. As we will eventually remove this part, looks okay to me now.

skyzh (approved):
LGTM if you don't want to include range filter in filter condition for now. Let's improve this in future PRs.

skyzh:
So to summarize, there are a list of things to do after this PR gets merged:

* Enforce pk not null, and pks are the first, second, etc., columns in binder. Alternatives: just ensure .iter()'s first column is `StorageColumnRef::Idx(x)`, where x is pk (can be done in optimizer).
* Include range filter in filter condition, and only pass approximate begin/end row id to RowSetIterator. The actual filter will be done by filter condition, and begin/end row id don't need to be accurate.
* Use [`partition_point`](https://doc.rust-lang.org/std/primitive.slice.html#method.partition_point) (or implement binary search by ourselves) when possible.

PR #646, opened 2022-05-14
feat: support sort agg

### LINKED ISSUE #591: executor: implement `SortAgg`
`SortAgg` is a special aggregator that takes an input of sorted group keys. For example:

```
select sum(x) from table group by group_key;

| group_key | x |
| a         | 1 |
| a         | 2 |
| a         | 3 |
| b         | 1 |
```

As data are already sorted with group_key, `SortAgg` can only have state for one group at a time, which significantly reduces memory footprint.

Kikkon:
close #591

### DIFF
--- a/src/executor/sort_agg.rs
+++ b/src/executor/sort_agg.rs
@@ -41,7 +41,7 @@ impl SortAggExecutor {
                 for col in group_cols.iter() {
                     group_key.push(col.get(row_idx));
                 }
-                // Check group key & lask key
+                // Check group key & last key
                 match last_key {
                     Some(last_key) => {
                         if last_key != group_key {
--- a/src/executor/sort_agg.rs
+++ b/src/executor/sort_agg.rs
@@ -1,0 +1,245 @@
+// Copyright 2022 RisingLight Project Authors. Licensed under Apache-2.0.
+#![allow(dead_code)]
+use smallvec::SmallVec;
+
+use super::*;
+use crate::array::{ArrayBuilderImpl, ArrayImpl};
+use crate::binder::BoundAggCall;
+use crate::types::{DataTypeExt, DataTypeKind};
+
+pub struct SortAggExecutor {
+    pub agg_calls: Vec<BoundAggCall>,
+    pub group_keys: Vec<BoundExpr>,
+    pub child: BoxedExecutor,
+}
+
+impl SortAggExecutor {
+    #[try_stream(boxed, ok = DataChunk, error = ExecutorError)]
+    pub async fn execute(self) {
+        let mut last_key = None::<HashKey>;
+        let mut states = create_agg_states(&self.agg_calls);
+
+        #[for_await]
+        for chunk in self.child {
+            // Eval group keys and arguments
+            let chunk = chunk?;
+            let exprs: SmallVec<[ArrayImpl; 16]> = self
+                .agg_calls
+                .iter()
+                .map(|agg| agg.args[0].eval(&chunk))
+                .try_collect()?;
+            let group_cols: SmallVec<[ArrayImpl; 16]> = self
+                .group_keys
+                .iter()
+                .map(|e| e.eval(&chunk))
+                .try_collect()?;
+
+            let num_rows = chunk.cardinality();
+            for row_idx in 0..num_rows {
+                // Create group key
+                let mut group_key = HashKey::new();
+                for col in group_cols.iter() {
+                    group_key.push(col.get(row_idx));
+                }
+                // Check group key & lask key
+                match last_key {
+                    Some(last_key) => {
+                        if last_key != group_key {
+                            yield Self::finish_agg(&states);
+                            states = create_agg_states(&self.agg_calls);
+                        }
+                    }
+                    None => (),
+                }
+                for (state, expr) in states.iter_mut().zip_eq(&exprs) {
+                    state.update_single(&expr.get(row_idx))?;
+                }
+                last_key = Some(group_key);
+            }
+        }
+        yield Self::finish_agg(&states);
+
+    }
+
+    fn finish_agg(states: &SmallVec<[Box<dyn AggregationState>; 16]>) -> DataChunk {
+        return states
+            .iter()
+            .map(|s| {
+                let result = &s.output();
+                match &result.data_type() {
+                    Some(r) => {
+                        let mut builder = ArrayBuilderImpl::with_capacity(1, r);
+                        builder.push(result);
+                        builder.finish()
+                    }
+                    None => ArrayBuilderImpl::new(&DataTypeKind::Int(None).nullable()).finish(),
+                }
+            })
+            .collect::<DataChunk>();
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+    use crate::array::ArrayImpl;
+    use crate::binder::{AggKind, BoundInputRef};
+    use crate::types::{DataType, DataTypeKind};
+
+    #[tokio::test]
+    async fn test_multi_group_agg() {
+        test_group_agg(
+            vec![0, 1],
+            vec![1, 2],
+            vec![
+                vec![1.1, 0.2, 0.3, 0.4, 0.5],
+                vec![1.1, 1.1, 1.3, 1.4, 1.5],
+                vec![1.2, 1.2, 2.3, 2.4, 2.5],
+            ],
+            vec![
+                vec![1.3, 2.2],
+                vec![0.3, 1.3],
+                vec![0.4, 1.4],
+                vec![0.5, 1.5],
+            ],
+        )
+        .await;
+        test_group_agg(
+            vec![0, 1],
+            vec![1, 2],
+            vec![
+                vec![0.1, 0.2, 0.3, 0.4, 0.5],
+                vec![1.1, 1.1, 1.3, 1.4, 1.5],
+                vec![1.1, 1.2, 2.3, 2.4, 2.5],
+            ],
+            vec![
+                vec![0.1, 1.1],
+                vec![0.2, 1.1],
+                vec![0.3, 1.3],
+                vec![0.4, 1.4],
+                vec![0.5, 1.5],
+            ],
+        )
+        .await
+    }
+
+    #[tokio::test]
+    async fn test_single_group_agg() {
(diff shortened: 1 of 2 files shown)


Kikkon:
please help review, and give me some more advice. 🤣

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,189 @@
...
+
+                    for (state, expr) in states.iter_mut().zip_eq(&exprs) {
+                        state.update_single(&expr.get(row_idx))?;
+                    }
+                } else {
+                    let mut states = create_agg_states(&self.agg_calls);
+                    for (state, expr) in states.iter_mut().zip_eq(&exprs) {
+                        state.update_single(&expr.get(row_idx))?;
+                    }
+                    states_vec.push(states);
```
xxchan, line 57: Why not yield the agg result of last group here? If so, there's no need to store all previous states.

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,189 @@
...
+        };
+        let mut executor = executor.execute();
+
+        let expected_col = vec![1.3, 2.2, 0.3, 1.3, 0.4, 1.4, 0.5, 1.5];
+        let mut result_col = Vec::new();
+        while let Some(chunk) = executor.next().await {
+            for chunk in chunk.unwrap().arrays() {
+                result_col.push(chunk.clone());
+            }
+        }
```
xxchan, line 126: This test code seems a little bit confusing. Suggest to represent `expected` as `vec![vec![1.3, 2.2], ...]` And then assert each chunk looks like that.

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,189 @@
+// Copyright 2022 RisingLight Project Authors. Licensed under Apache-2.0.
+#![allow(dead_code)]
+use itertools::Itertools;
```
xxchan, line 3: Can it be removed?

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub agg_calls: Vec<BoundAggCall>,
+    pub group_keys: Vec<BoundExpr>,
+    pub child: BoxedExecutor,
+}
+
+impl SortAggExecutor {
+    #[try_stream(boxed, ok = DataChunk, error = ExecutorError)]
+    pub async fn execute(self) {
+        let mut states_vec: Vec<SmallVec<[Box<dyn AggregationState>; 16]>> = vec![];
+        let mut last_key = Arc::new(HashKey::new());
```
xxchan, line 21: Is the `Arc` unnecessary?

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,189 @@
...
+
+            let num_rows = chunk.cardinality();
+            for row_idx in 0..num_rows {
+                // Create group key
+                let mut group_key = HashKey::new();
+                for col in group_cols.iter() {
+                    group_key.push(col.get(row_idx));
+                }
+                let group_key = Arc::new(group_key);
+                if last_key.len() != 0 && group_key.eq(&last_key) {
```
xxchan, line 46: I suggest to use `Option<HashKey>` instead of `last_key.len() != 0`

xxchan (commented):
Please ping me again when you are ready for review. Also feel free to ask for help about specific questions. 😊

src/executor/sort_agg.rs
Kikkon: When you have time, hope to help review again. I'm not familiar with async in rust, if there are mistakes, please help point them out (by the way, there are related materials that can also be recommended.  (：thx

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,238 @@
...
+                .map(|agg| agg.args[0].eval(&chunk))
+                .try_collect()?;
+            let group_cols: SmallVec<[ArrayImpl; 16]> = self
+                .group_keys
+                .iter()
+                .map(|e| e.eval(&chunk))
+                .try_collect()?;
+
+            let num_rows = chunk.cardinality();
+            let mut states = create_agg_states(&self.agg_calls);
```
xxchan, line 37: Should this variable be defined outside the `for chunk in self.child` loop?

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,238 @@
...
+            for row_idx in 0..num_rows {
+                // Create group key
+                let mut group_key = HashKey::new();
+                for col in group_cols.iter() {
+                    group_key.push(col.get(row_idx));
+                }
+                // Check group key & lask key
+                if last_key.is_some() && group_key.eq(last_key.as_ref().unwrap()) {
```
xxchan, line 45: `is_some` + `unwrap` (2 operations) can be simplified as `if let Some(..) = opt` or `match opt` (1 operation).

src/executor/sort_agg.rs
xxchan: Be confident!

src/executor/sort_agg.rs
xxchan: e.g., child may have two chunks `[1,1], [1,2]`

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,238 @@
...
+pub struct SortAggExecutor {
+    pub agg_calls: Vec<BoundAggCall>,
+    pub group_keys: Vec<BoundExpr>,
+    pub child: BoxedExecutor,
+}
+
+impl SortAggExecutor {
+    #[try_stream(boxed, ok = DataChunk, error = ExecutorError)]
+    pub async fn execute(self) {
+        let mut last_key = None::<SmallVec<[DataValue; 16]>>;
```
xxchan, line 19:
```suggestion
        let mut last_key = None::<HashKey>;
```

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,238 @@
...
+                        state.update_single(&expr.get(row_idx))?;
+                    }
+                } else {
+                    if last_key.is_some() {
+                        yield finish_agg(states);
+                        states = create_agg_states(&self.agg_calls);
+                    }
+                    for (state, expr) in states.iter_mut().zip_eq(&exprs) {
+                        state.update_single(&expr.get(row_idx))?;
+                    }
```
xxchan, line 56: Please consider again the branching logic here. Actually you only need one `if`, and then update state outside if

```rust 
if new agg key occurs {
    yield current state & reset state
}
update state acoording to the row
```

src/executor/sort_agg.rs
xxchan: BTW, you don't need to be familiar with async actually. You can just "feel it". 😄

Specifically, to finish this work you just need to know the **semantics** of `yield` and generators.

src/executor/sort_agg.rs
Kikkon: You mean [1,1] [1,2], group key in index[0], the result is [3,2]？

xxchan: Yes

xxchan (commented):
Others LGTM! Ping @skyzh to take a look.

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,244 @@
...
+                        }
+                    }
+                    None => (),
+                }
+                for (state, expr) in states.iter_mut().zip_eq(&exprs) {
+                    state.update_single(&expr.get(row_idx))?;
+                }
+                last_key = Some(group_key);
+            }
+            yield finish_agg(&states)
```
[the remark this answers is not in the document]
xxchan: So I think adding relevant tests may be better..

xxchan:
BTW, since we will squash merge, there's no need to force push everytime to make commit history (too) clean. Leave changes in new commits may also help reviewers to focus on new changes since last review 😄

skyzh (commented):
Would you please add a test case when there are no rows from child? We should output no rows instead of one row of all null. I guess the current behavior is incorrect 🤣

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,244 @@
...
+        for index in agg_call_index {
+            agg_calls.push(create_sum_agg_call(index));
+        }
+
+        let mut group_keys = vec![];
+        for index in group_key_index {
+            group_keys.push(create_input_ref(index));
+        }
+
+        let child: BoxedExecutor = async_stream::try_stream! {
```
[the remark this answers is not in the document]
Kikkon: SimpleAggExecutor has the same problem too?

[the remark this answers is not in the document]
Kikkon: get it, thx!

Kikkon:
@xxchan @skyzh  any other suggestions? 😄

skyzh:
Let me take a look 🤣

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,251 @@
...
+                .try_collect()?;
+
+            let num_rows = chunk.cardinality();
+            for row_idx in 0..num_rows {
+                // Create group key
+                let mut group_key = HashKey::new();
+                for col in group_cols.iter() {
+                    group_key.push(col.get(row_idx));
+                }
+                // Check group key & lask key
```
skyzh, line 44: lask -> last

src/executor/sort_agg.rs
```diff
@@ -0,0 +1,251 @@
...
+                .iter()
+                .map(|e| e.eval(&chunk))
+                .try_collect()?;
+
+            let num_rows = chunk.cardinality();
+            for row_idx in 0..num_rows {
+                // Create group key
+                let mut group_key = HashKey::new();
+                for col in group_cols.iter() {
+                    group_key.push(col.get(row_idx));
```
skyzh, line 42: Maybe we can optimize this in the future. We can first check group key is the same as the current group key, and then clone by `.get` if different. Otherwise we will always need to create new vector for every row of hash key.

skyzh (approved):
Rest LGTM, thanks! Would you please help fix the typo?

src/executor/sort_agg.rs
Kikkon: sure.

Kikkon:
Thank you for your support 😊  @xxchan @skyzh

PR #647, opened 2022-05-15
feat(optimizer): support (filter_agg_rule + having clause)

lokax:
**Support having clause and alias**
```sql
select count(x) as a, y as b from test group by b having a > 1

select 42 having 42 > 18;
```
**Forbid to generate:**
```sql
select count(x) from test group by count(x)
```

**Support push the filter down below the agg**
**Before:**
```text
> explain select count(x), y + 1 from test group by y + 1 having count(x) > 1 and y + 1 = 2;
PhysicalProjection:
    InputRef #1
    InputRef #0
  PhysicalFilter: expr And(Gt(InputRef #1, Int32(1) (const)), Eq(InputRef #0, Int32(2) (const)))
    PhysicalHashAgg:
        (InputRef #1 + 1)
        count(InputRef #0) -> INT
      PhysicalTableScan:
          table #13,
          columns [0, 1],
          with_row_handler: false,
          is_sorted: false,
          expr: None
``` 
**After:**
```text
> explain select count(x), y + 1 from test group by y + 1 having count(x) > 1 and y + 1 = 2;
PhysicalProjection:
    InputRef #1
    InputRef #0
  PhysicalFilter: expr Gt(InputRef #1, Int32(1) (const))
    PhysicalHashAgg:
        (InputRef #1 + 1)
        count(InputRef #0) -> INT
      PhysicalTableScan:
          table #13,
          columns [0, 1],
          with_row_handler: false,
          is_sorted: false,
          expr: Eq(Plus(InputRef #1, Int32(1) (const)), Int32(2) (const))
```

### DIFF
--- a/src/binder/expression/column_ref.rs
+++ b/src/binder/expression/column_ref.rs
@@ -90,9 +90,15 @@ impl Binder {
                 }
             }
             if info == None {
-                if self.context.aliases.contains(column_name) {
+                if let Some(index) = self
+                    .context
+                    .aliases
+                    .iter()
+                    .position(|name| column_name == name)
+                {
                     Ok(BoundExpr::Alias(BoundAlias {
                         alias: column_name.clone(),
+                        return_type: self.context.aliases_expr_return_type[index].clone(),
                     }))
                 } else {
                     Err(BindError::InvalidColumn(column_name.clone()))
--- a/src/binder/expression/expr_with_alias.rs
+++ b/src/binder/expression/expr_with_alias.rs
@@ -27,6 +27,8 @@ impl std::fmt::Display for BoundExprWithAlias {
 #[derive(PartialEq, Clone, Serialize)]
 pub struct BoundAlias {
     pub alias: String,
+    pub return_type: Option<DataType>,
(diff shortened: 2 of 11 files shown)


xxchan (commented):
I think filter_agg_rule and having clause can be splited into 2 atomic PRs. Or do you have some special reasons to implement them together?

lokax:
> I think filter_agg_rule and having clause can be splited into 2 atomic PRs. Or do you have some special reasons to implement them together?
No😑 
Just for testing.
Because the having clause can create a filter above the agg operator.

lokax:
> I think filter_agg_rule and having clause can be splited into 2 atomic PRs. Or do you have some special reasons to implement them together?

No😑
Just for testing.
Because the having clause can create a filter above the agg operator.

xxchan:
Oh yes it makes sense since we don't have subquery yet 😇

xxchan (commented):
The rule LGTM. But I haven't looked at the implementation of `having` yet, and it seems non-trivial. Could you please create a new PR for `having` and briefly explain the ideas of the implementation? 😊

PR #649, opened 2022-05-17
binder: primary key should by-default be non-null

### LINKED ISSUE #632: binder: primary key should by-default be non-null
```sql
create table t(a int primary key, b int);
```
will produce a  `BoundCreateTable` like this 
```
BoundCreateTable {
        database_id: 0,
        schema_id: 0,
        table_name: "t",
        columns: [
            ColumnCatalog {
                id: 0,
                desc: ColumnDesc {
                    datatype: Int(None) (null),
                    name: "a",
                    is_primary: true,
                },
            },
            ColumnCatalog {
                id: 1,
                desc: ColumnDesc {
                    datatype: Int(None) (null),
                    name: "b",
                    is_primary: false,
                },
            },
        ],
    },
```
I think we need to make primary key default not to be null, so if i type sql query above, we will make `a` not null. Or we can forbid that syntax.

the reason for this is that
https://github.com/risinglightdb/risinglight/blob/d28389a62069cee31882dcae4a4e10383e4518f2/src/binder/statement/create_table.rs#L90-L99
when we only type `a int primary key`, we will go to line 98 directly without setting `is_nullable` to false

Ted-Jiang:
This is my first pr to this repo 😁，If there are something i miss plz fell free to tell me.
close #632

### DIFF
--- a/src/binder/statement/create_table.rs
+++ b/src/binder/statement/create_table.rs
@@ -77,6 +77,7 @@ impl Binder {
                 // // Remove this line and change `columns` above to immut.
                 for &index in &ordered_pk_ids {
                     columns[index as usize].set_primary(true);
+                    columns[index as usize].set_nullable(false);
                 }
 
                 Ok(BoundCreateTable {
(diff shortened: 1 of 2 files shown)


skyzh:
cc @Shmiwy now the pk won't be nullable any more. I guess this would make things easier on storage layer.

PR #651, opened 2022-05-17
feat(binder): support Having clause and more flexible OrderBy

lokax:
Signed-off-by: lokax <m632656684@gmail.com>
### Having clause
**Support having clause and alias**
```sql
select count(x) as a, y as b from test group by b having a > 1

select 42 having 42 > 18;
```
**Forbid to generate:**
```sql
select count(x) from test group by count(x)
```

**Since the above of `Agg` is not necessarily a `Project`, a more general way to generate binding is needed.**
**I also push `AggCall` Expression into bingding vector in the `InputRefResolver`**

**The search process:**
Step 1. Find binding for y + 1 expression
```sql
select y + 1 from test group by y + 1
```
Step 2. If not found: Find binding for y expression
```sql
select y + 1 from test group by y
```
-------------------------------------------------------------------------------
### Problems with aliases
```sql
select x as i from test group by x having i + 1
```
Because the return type of child needs to be known when the `BinaryOp(i + 1)` expression is bound.
```rust
pub struct BoundAlias {
    pub alias: String,
    pub expr: Box<BoundExpr>, // I add a new field
}
```
`AliasRewriter` will rewrite alias expressions when generating the logical plan

-------------------------------------------------------------------------------
### More flexible OrderBy
```sql
select x from test order by y
```
If the expression of `OrderBy` does not exist in the select list, an expression copy will be added to the select list to generate the correct binding.

**Change: select list: [column(x)] -> [column(x), column(y)]**

**If the select list has changed:**
Also add an new `Project` above the `OrderBy` to produce the correct output: [column(x), column(y)] -> [column(x)]

Now `OrderBy` can support sort using expressions
```sql
select x from test order by -(x + 1)
select y from test group by y order by count(x);
```


skyzh (approved):
LGTM! For unused code, maybe we can simply remove them?

Issue #660, opened 2022-06-09
storage: abstract disk I/O


skyzh:
Currently, we have create_dir, etc. everywhere. We'd better have a single interface to operate files on disk, so as to support in-memory / disk / object store backends.

wangrunji0408:
Isn't that only the disk backend depends on file operations?

skyzh:
Yes. But even disk backend has two modes -- pure in-memory mode (for testing, where files are stored in a hash map), and real on-disk mode.

skyzh:
If we can abstract all disk operations to use a trait like ObjectStore, we can prevent unwanted writes to disk.

wangrunji0408:
I think we should use the real on-disk mode in testing, to make sure we are correctly using the system fs API. Data path can be redirected to ramfs/tmpfs to speed up.

wangrunji0408:
Later we can also introduce the simulation testing, where all fs API will be mocked to a in-memory simulator.

skyzh:
I switch to pure in-memory mode because tmpfs is slow 🤣 fsync and manifest write will add 1s latency to every disk test case, which looks weird.

Maybe I can try switch off fsync for all writes (even manifest), and maybe things will work.

wangrunji0408:
🤣 okay let's stay in memory mode for efficiency.

PR #661, opened 2022-06-09
feat(test): add sqlplannertest

skyzh:

As title. It's a really big change :)

Basically, we will need to run `make apply_planner_test` and add the latest planner test result into the git repo.

This PR also fixes some issues about disk I/O in in-memory mode. https://github.com/risinglightdb/risinglight/issues/660 should further fix this.

### DIFF
--- a/docs/05-e2e-tests.md
+++ b/docs/05-e2e-tests.md
@@ -44,7 +44,7 @@ The test cases are stored in `tests/planner_test`. Use the following command:
 make apply_planner_test
 ```
 
-to automatically update the test result.
+to generate a sql file containing explain results for each yaml file.
 
 Generally, we will compare the explain result before and after a commit, so as to know how the commit
 affects the optimizer result. We don't really care about the correctness -- we just compare the explain
--- a/docs/05-e2e-tests.md
+++ b/docs/05-e2e-tests.md
@@ -1,0 +1,51 @@
+# SQLLogicTest and SQLPlannerTest
+
+RisingLight uses two test frameworks to do end-to-end tests.
+
+## SQLLogicTest
+
+SQLLogicTest will run a special `slt` file and compare the result from the expected output.
+The test cases are stored under `tests/sql` folder.
+
+For example, let's see `order_by.slt`:
+
+```
+statement ok
+create table t(v1 int not null, v2 int not null)
+
+statement ok
+insert into t values(1, 1), (4, 2), (3, 3), (10, 12), (2, 5)
+
+query I
+select v1 from t order by v1 asc
+----
+1
+2
+3
+4
+10
+```
+
+The first 3 test cases of this test file are
+* check whether create table works
+* check whether insert table works 
+* select data from table
+
+We use `statement ok` to ensure statement successfully runs, and `query` to compare the query result.
+
+When running `make test`, the test runner will run all files under `tests/sql` folder.
+
+## SQLPlannerTest
+
+SQLPlannerTest is a regression test. We will write yaml files to describe the cases we want to test.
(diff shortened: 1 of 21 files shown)


tests/sqllogictest/src/lib.rs
```diff
@@ -28,12 +27,7 @@ pub async fn test_mem(name: &str) {
 
 pub async fn test_disk(name: &str) {
     init_logger();
-    // still need a temp_dir to write manifest
-    let temp_dir = tempdir().unwrap();
```
wangrunji0408, line 32: So now testing with the secondary storage will not actually touch files?

docs/05-e2e-tests.md
```diff
@@ -0,0 +1,51 @@
...
+## SQLPlannerTest
+
+SQLPlannerTest is a regression test. We will write yaml files to describe the cases we want to test.
+The test cases are stored in `tests/planner_test`. Use the following command:
+
+```
+make apply_planner_test
+```
+
+to automatically update the test result.
```
wangrunji0408, line 47: To clarify the relationship between yml and sql files.

```suggestion
to generate a sql file containing explain results for each yaml file.
```

tests/planner_test/count.yml
```diff
@@ -0,0 +1,14 @@
+- sql: |
+    explain select count(*) from t
+  desc: count(*) is special
+  before:
+    - create table t(v int)
+  tasks:
+    - print
+- sql: |
```
wangrunji0408, line 8:
```suggestion

- sql: |
```

wangrunji0408 (commented):
LGTM!
I didn't review the storage part. Not very familiar to me. 😇
And I'm still wondering if it's possible to integrate sqlplannertest into sqllogictest.

tests/sqlplannertest/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "risinglight_plannertest"
+version = "0.1.0"
+edition = "2021"
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = { version = "1", features = ["backtrace"] }
+async-trait = "0.1"
+risinglight = { path = "../.." }
+sqlplannertest = { git = "https://github.com/risinglightdb/sqlplannertest-rs" }
```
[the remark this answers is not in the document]
skyzh: We have Cargo.lock, so I don't think we'll need to lock version.

tests/sqllogictest/src/lib.rs
skyzh: Yes, I've manually removed all disk access. I think we will eventually have a object store abstraction in the future.

wangrunji0408: It sounds unreasonable to test a disk backend without actually accessing files.

skyzh: Actually, the reason is that, we only test whether the disk encoder / decoder can correctly read the file format, without actually testing things like "fsync", especially for unit tests like sqllogictest.

wangrunji0408: I still feel that sqllogictest is a kind of e2e test and it's better to cover the real execution path as much as possible. 🤔

skyzh: If we can abstract file operations into a single trait, then all paths are still covered, except the on-disk implementation of ObjectStore or FileSystem trait. Indeed RisingLight currently do not rely much on the filesystem magics and properties, so I don't think it will be necessary to cover that paths.

PR #662, opened 2022-06-22
feat(cli): support multi-line interactive mode

Gun9niR:
Signed-off-by: Gun9niR <gun9nir.guo@gmail.com>

https://github.com/risinglightdb/risinglight/issues/475

`;` at the end of a line terminates input for the current statement(s).

PR #664, opened 2022-06-25
fix: 'unsupported command' everywhere

cadl:
Signed-off-by: cadl <ctrlaltdeleteliu@gmail.com>

After #662, the internal command does not work. Make some fixes.

before:
```
> \dt
? ;
Internal error: unsupported command
> \dt;
Internal error: unsupported command
>
```

after:
```
> \dt
+---+----------+---+------------+---+--------------+
| 0 | postgres | 1 | pg_catalog | 0 | contributors |
+---+----------+---+------------+---+--------------+
in 0.001s
> select * from foo;
bind error: invalid table foo
> select *
? from foo;
bind error: invalid table foo
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -154,6 +154,12 @@ fn read_sql(rl: &mut Editor<()>) -> Result<String, ReadlineError> {
         if line.is_empty() {
             continue;
         }
+
+        // internal commands starts with "\"
+        if line.starts_with('\\') && sql.is_empty() {
+            return Ok(line);
+        }
+
         sql.push_str(line.as_str());
         if line.ends_with(';') {
             return Ok(sql);


TennyZhuang (approved):
Thanks!

skyzh (approved):
Thanks! In fact, I think `\dt` should always work without `;`. That is to say, if a line begins with `\`, then the cli should immediately execute it on new line instead of waiting for `;`. If you're interested, you may submit another PR to fix the behavior.

cadl:
> Thanks! In fact, I think `\dt` should always work without `;`. That is to say, if a line begins with `\`, then the cli should immediately execute it on new line instead of waiting for `;`. If you're interested, you may submit another PR to fix the behavior.

yes, `\dt` works like you say now.  As the example in the PR description

```
> \dt
+---+----------+---+------------+---+--------------+
| 0 | postgres | 1 | pg_catalog | 0 | contributors |
+---+----------+---+------------+---+--------------+
in 0.001s
```

skyzh:
That's cool, thanks!

Issue #531, opened 2022-02-24
storage: using variable size encoding for RLE count


ludics:
Currently, we use u16 for the count of run-length encoding:
https://github.com/risinglightdb/risinglight/blob/613ffb6990458bb3558b923ce77ea1e3c1dced4a/src/storage/secondary/block/rle_block_builder.rs#L16-L26

Also we can consider using variable size encoding (IIRC, prost::encode_varint?) for those integers, and therefore we can store elements more efficiently, and we don't need to limit the RLE count to u16. This could be done in the future; I'm not sure if we can optimize storage space by using variable size encoding for RLE count.

_Originally posted by @skyzh in https://github.com/risinglightdb/risinglight/pull/507#discussion_r811545436_

unconsolable:
Seems it has been resolved by #666 @skyzh

PR #666, opened 2022-07-11
feat(storage): use variable size encoding to compress rle count

xinchengforgit:
Signed-off-by: xinchengforgit 2419743144@qq.com

This PR try to solve #531.
I have some basic thoughts about variable size encoding, but there are some tests that I don't know how to modify, please give me some suggestions.

I use variable size encoding, and replace the u16 with u32, so that we can support a larger range number while saving the used space.

### DIFF
--- a/src/storage/secondary/block/rle_block_builder.rs
+++ b/src/storage/secondary/block/rle_block_builder.rs
@@ -59,19 +59,19 @@ pub fn decode_u32_slice(bytes: &[u8]) -> Result<(u32, usize), DecodeError> {
     b = unsafe { *bytes.get_unchecked(1) };
     part0 += u32::from(b) << 7;
     if b < 0x80 {
-        return Ok((u32::from(part0), 2));
+        return Ok((part0, 2));
     };
     part0 -= 0x80 << 7;
     b = unsafe { *bytes.get_unchecked(2) };
     part0 += u32::from(b) << 14;
     if b < 0x80 {
-        return Ok((u32::from(part0), 3));
+        return Ok((part0, 3));
     };
     part0 -= 0x80 << 14;
     b = unsafe { *bytes.get_unchecked(3) };
     part0 += u32::from(b) << 21;
     if b < 0x80 {
-        return Ok((u32::from(part0), 4));
+        return Ok((part0, 4));
     };
     part0 -= 0x80 << 21;
     b = unsafe { *bytes.get_unchecked(4) };
--- a/src/storage/secondary/block/rle_block_builder.rs
+++ b/src/storage/secondary/block/rle_block_builder.rs
@@ -104,7 +104,7 @@ impl<A, B> RleBlockBuilder<A, B>
 where
     A: Array,
     B: BlockBuilder<A>,
-    A::Item: PartialEq, // 可以实现== 比较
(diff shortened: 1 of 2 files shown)


src/storage/secondary/block/rle_block_builder.rs
```diff
@@ -2,34 +2,109 @@
...
 
-use bytes::BufMut;
+use bytes::{Buf, BufMut};
+use prost::DecodeError;
 use risinglight_proto::rowset::BlockStatistics;
 
 use super::BlockBuilder;
 use crate::array::Array;
 
+pub fn encode_32<B>(mut value: u32, buf: &mut B)
```
skyzh, line 12: Would you please write docstring for this function? Also, would you please explain a little bit about the implementation? (e.g. borrowed from other products? Or some general ideas about this)

src/storage/secondary/block/rle_block_builder.rs
```diff
@@ -166,7 +246,7 @@ mod tests {
             .iter()
             .cycle()
             .cloned()
-            .take(u16::MAX as usize * 2)
+            .take(u32::MAX as usize + 1)
```
skyzh, line 249: Will this cause test to be running very slow? If so, we can avoid testing it.

src/storage/secondary/block/rle_block_builder.rs
```diff
@@ -88,9 +163,14 @@ where
         }
         self.rle_counts.push(self.cur_count);
         encoded_data.put_u32_le(self.rle_counts.len() as u32);
+        let offset = encoded_data.len();
+        encoded_data.put_u32_le(0xffffffff); // placeholder
```
skyzh, line 167: Do we really need this length? We can read until the end when decoding given the number of RLE count item.

skyzh (commented):
Rest LGTM, good work! Please help fix fmt. See developer docs for more information.

src/storage/secondary/block/rle_block_builder.rs
xinchengforgit: Yeah, In fact, I borrow the idea from the prost::encode_varint.
This function encodes u32 to multiple u8 , and for every u8, the highest bit tells whether this u8 belongs to the same u32.

src/storage/secondary/block/rle_block_builder.rs
xinchengforgit: I think it does cause test to be running very slow. But I don't know how to change the test myself.

src/storage/secondary/block/rle_block_builder.rs
skyzh: Please add license header of prost in this file.

src/storage/secondary/block/rle_block_builder.rs
```diff
@@ -20,9 +94,11 @@ where
     A::Item: PartialEq,
 {
     block_builder: B,
-    rle_counts: Vec<u16>,
+    rle_counts: Vec<u32>,
```
skyzh, line 97: Please update comments on L88

skyzh (approved):
Rest LGTM, good work!

PR #667, opened 2022-07-13
fix: create table with invalid primary keys constraints

GoGim1:
CreateTable's statement should raise an error, when its constraints contain invalid primary keys.

### DIFF
--- a/src/binder/statement/create_table.rs
+++ b/src/binder/statement/create_table.rs
@@ -148,7 +148,7 @@ impl Binder {
                     columns,
                     ..
                 } if *is_primary => columns.iter().for_each(|ident| {
-                    pks_name_from_constraints.push(ident.value.to_lowercase());
+                    pks_name_from_constraints.push(ident.value.clone());
                 }),
                 _ => continue,
             }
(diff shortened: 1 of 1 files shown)


src/binder/statement/create_table.rs
```diff
@@ -143,7 +148,7 @@ impl Binder {
                     columns,
                     ..
                 } if *is_primary => columns.iter().for_each(|ident| {
-                    pks_name_from_constraints.push(ident.value.clone());
+                    pks_name_from_constraints.push(ident.value.to_lowercase());
```
skyzh, line 151: I think we can keep everything case-sensitive for now. I'd prefer not to convert to lowercase.

skyzh (commented):
Rest LGTM. Please sign DCO following the developer guide, thanks!

src/binder/statement/create_table.rs
GoGim1: done

GoGim1: done

skyzh (approved):

PR #680, opened 2022-08-06
chore: add github codespace devcontainer configuration

yeya24:
This PR adds support for setting up the dev environment for Risinglight with Github Codespace. https://docs.github.com/en/codespaces


skyzh:
Thanks for this PR! But I doubt whether this is necessary. Do you mainly develop RisingLight on GitHub Codespace?

yeya24:
This comes in handy sometimes to start a dev environment for Rust and Risinglight to read/write the code. But it is not as good as the local environment due to the resource limitation, which makes it takes a long time to compile the code.

I think it is a good compliment because it gives you an out-of-box environment with the required deps. For beginners like me this is already very useful to try the project.

skyzh:
That's cool.

* I don't think it's good to use zsh here. We should keep the environment as simple as possible.
* Should install the Rust version of `rust-toolchain` file in RisingLight repo, instead of just specifying nightly toolchain.

yeya24:
Thanks. I have updated the PR, please take a look.

skyzh (approved):

PR #684, opened 2022-08-22
fix:make test

Wambug:
Signed-off-by: wambug <wambugumacharia35@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ build:
 	cargo build --all-features --all-targets
 
 test:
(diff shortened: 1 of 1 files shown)


xxchan:
Thanks! But we are using `nextest` to run tests. https://github.com/risinglightdb/risinglight/blob/main/CONTRIBUTING.md#write-tests

You'll need `cargo install cargo-nextest` to run tests.

Wambug:
ok 😅

PR #685, opened 2022-08-25
storage: use Arc<Version> during pin and unpin

chaixuqing:
Signed-off-by: chaixuqing [xuqingchai00@gmail.com](mailto:xuqingchai00@gmail.com)

close https://github.com/risinglightdb/risinglight/issues/537

Currently, we use `pin` and `unpin` manually in code. In this request, we use `Arc<Version>` to help us automatically unpin instead of manual code.

### DIFF
--- a/src/storage/secondary/compactor.rs
+++ b/src/storage/secondary/compactor.rs
@@ -179,14 +179,14 @@ impl Compactor {
         loop {
             {
                 let tables = self.storage.tables.read().clone();
-                let (epoch, snapshot) = self.storage.version.pin();
+                let pin_version = self.storage.version.pin();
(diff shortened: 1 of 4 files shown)


skyzh (approved):
LGTM. Still the other idea mentioned in https://github.com/risinglightdb/risinglight/pull/549

> I've another idea: we can simply using Arc to help us do reference counting, instead of maintaining the ref_cnt by ourselves. For each epoch, we associate a single Version object, and wrap it with an Arc. When a user requests pin, we clone the Arc for them. When the last Arc gets dropped, the Version object gets dropped, and we can safely remove the version.

If you're interested, welcome to implement pin / unpin in this way.

PR #686, opened 2022-08-25
feat(planner): plan select distinct

### LINKED ISSUE #687: Executior: Suport First/Last aggregate function

### LINKED ISSUE #688: planner: Support select distinct

lokax:
Use HashAggregate to implement DISTINCT. The result will be unordered
Closes #687 
Closes #688


skyzh (approved):
Thanks!

tests/sql/distinct.slt
```diff
@@ -0,0 +1,19 @@
...
+
+query II
+SELECT DISTINCT * FROM test ORDER BY x;
+----
+1 1
+2 2
+3 3
+
+query I
+SELECT DISTINCT x FROM test ORDER BY y DESC;
```
xxchan, line 15: This is problematic. (Should be banned) e.g.,

```sql
> CREATE TABLE test (x INT, y INT);
created
in 0.001s
> insert into test values (1,3),(2,2),(3,1),(1,4);
4 rows inserted
in 0.001s
> select distinct x from test order by y;
+---+
| x |
+---+
| 3 |
| 2 |
| 1 |
| 1 |
+---+
in 0.003s
```

In postgres:
```sql
postgres=# SELECT DISTINCT x FROM test ORDER BY y DESC;
ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
LINE 1: SELECT DISTINCT x FROM test ORDER BY y DESC;
  
```

src/logical_planner/select.rs
```diff
@@ -135,7 +135,10 @@ impl LogicalPlaner {
         let comparators = stmt.orderby;
 
         // TODO: support the following clauses
-        assert!(!stmt.select_distinct, "TODO: plan distinct");
+        // assert!(!stmt.select_distinct, "TODO: plan distinct");
+        if stmt.select_distinct {
+            plan = Arc::new(LogicalDistinct::new(plan))
```
xxchan, line 140: What about directly create `LogicalAgg` here?

tests/sql/distinct.slt
lokax: We don't need to ban it. Because we will add `Y` to the select_lists in binder. The previous implementation did have a bug. Now, the initial select list will be grouped(column x), and a `first` aggregation function will be added to the Y(added by OrderBy).

src/logical_planner/select.rs
lokax: > What about directly creating `LogicalAgg` here?

LogicalAggregate will create a new column bindings, but LogicalDistinct will not. This is important for `SELECT DISTINCT x FROM test ORDER BY x`. 

LogicalOrderBy is after LogicalDistinct. 

LogicalOrder expr is ColumnRef(x).
LogicalDistinct expr is InputRef(0).
LogicalProjection expr is ColumnRef(x).

Because LogicalDistinct does not generate new bindings, this is no problem, because Order::ColumnRef(x) = Proj::ColumnRef(x), we can successfully create InputRef for LogicalOrder.

But, LogicalAggregate will generate new bindings which causes bindings to become [ InputRef(0) ]
Order::ColumnRef(x)  != Agg::InputRef(0)

xxchan: Sorry but I didn't get your point. IIUC `SELECT DISTINCT x FROM test ORDER BY x` should be equivalent to `SELECT x FROM test GROUP BY x ORDER BY x`, and it seems we can correctly handle the latter one. Why the column binding for Order would become a problem?

tests/sql/distinct.slt
xxchan: Is there any DB support this behavior? IMO it's an unusal behavior, and is like select a column not in `GROUP BY`, which should be explicitly and carefully used when needed. So we should ban it :)

xxchan: I found PG has a special syntax [DISTINCT ON](https://www.postgresql.org/docs/current/sql-select.html#SQL-DISTINCT) for this behavior. Anyway, I would say this fancy feature is completely optional and we don't need to support it in the first place.

skyzh (approved):

skyzh:
After some weeks' experiment with BusTub, I think we. should refactor RisingLight's binder + planner 🤣 I might work on this after this semester.

Issue #638, opened 2022-04-24
storage: select compaction strategy using statistics


skyzh:
If we find distinct value statistics is very small compared with row counts, we can use RLE encoding when building RowSets.

https://github.com/risinglightdb/risinglight/blob/65372ee4da36bada718a6ab00cdba865ba554e5e/src/storage/secondary/compactor.rs#L93-L96

jess-x:
newbie here, can I get this a try?

noneback:
Any progress? I am also very interested in it.

skyzh:
Hmm, maybe you can take over this as it has been nearly a month that we haven't heard back from @jess-x 🤣

jess-x:
hi, feel free to take over. didn't have the bandwidth this past month 🥲

Issue #670, opened 2022-07-20
expr: use `ordered_float` for all occurrences of `f64`


skyzh:
`f64` is not Hash and Eq, which might cause issues for some implementations within the system.

Kikkon:
I'm learning type exercise 🤣 I want to try this

PR #689, opened 2022-09-07
 chore(storage): select compaction strategy using statistics

chaixuqing:
Signed-off-by: chaixuqing [xuqingchai00@gmail.com](mailto:xuqingchai00@gmail.com)

close https://github.com/risinglightdb/risinglight/issues/638

If we find distinct value statistics is very small compared with row counts, we can use RLE encoding when building RowSets.

### DIFF
--- a/src/storage/secondary/compactor.rs
+++ b/src/storage/secondary/compactor.rs
@@ -4,6 +4,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use itertools::Itertools;
+use risinglight_proto::rowset::block_statistics::BlockStatisticsType;
 use tokio::sync::oneshot::Receiver;
 use tracing::{info, warn};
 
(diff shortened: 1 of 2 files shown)


src/storage/secondary/compactor.rs
```diff
@@ -97,10 +100,44 @@ impl Compactor {
...
+
+        let mut builder = if distinct_value < row_count / 5 {
+            RowsetBuilder::new(
+                table.columns.clone(),
+                ColumnBuilderOptions::from_storage_options(&table.storage_options),
+            )
+        } else {
+            let mut column_options =
+                ColumnBuilderOptions::from_storage_options(&table.storage_options);
+            column_options.encode_type = EncodeType::RunLength;
```
skyzh, line 138: We can use `DictEncoding`. If there're few distinct values but the data are like `1,2,3,1,2,3,1,2,3,...`, RLE might perform badly.

src/storage/secondary/rowset/disk_rowset.rs
```diff
@@ -109,6 +109,10 @@ impl DiskRowset {
         self.columns[storage_column_id].clone()
     }
 
+    pub fn get_columns(&self) -> Vec<Column> {
```
skyzh, line 112: Can we return `&[Column]`?

skyzh (commented):
Rest LGTM, good work!

src/storage/secondary/compactor.rs
chaixuqing: Dict encoding for f64 is not supported yet. I couldn't pass risinglight_sqllogictest::sqllogictest disk::limit.slt

skyzh: Makes sense... https://github.com/risinglightdb/risinglight/issues/670

Then let's try RLE first.

skyzh (approved):

PR #692, opened 2022-09-28
fix: add row_handler to schema

### LINKED ISSUE #450: optimizer: table scan schema should include row handler
Currently, the implementation of TableScan's schema is not correct.

https://github.com/risinglightdb/risinglight/blob/c221b085ceb4a61fe1cf1e6c9d6caa36b7215a2e/src/optimizer/plan_nodes/logical_table_scan.rs#L75-L77

When `with_row_handler` is enabled, we should have an extra i64 column for row handlers at the end.

yinfredyue:
This PR close #450.

Signed-off-by: Yue Yin <yueyin.dev@gmail.com>

### DIFF
--- a/src/optimizer/plan_nodes/logical_table_scan.rs
+++ b/src/optimizer/plan_nodes/logical_table_scan.rs
@@ -5,6 +5,7 @@ use std::fmt;
 
 use itertools::Itertools;
 use serde::Serialize;
+pub use sqlparser::ast::DataType as DataTypeKind;
 
 use super::*;
 use crate::binder::ExprVisitor;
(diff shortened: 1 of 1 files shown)


skyzh (approved):

skyzh:
Indeed this is not always correct -- row handler will only be in TableScan when there's update / delete. We can fix this when we decide on a rewrite of binder / planner / optimizer.

yinfredyue:
> Indeed this is not always correct -- row handler will only be in TableScan when there's update / delete. We can fix this when we decide on a rewrite of binder / planner / optimizer.

You are right. I did not fully understand how the row handlers are created/used when I was creating this PR. 

> row handler will only be in TableScan when there's update / delete. 

I just read the planner & executor code and this is indeed the case. However, what's the reason for creating this extra temporary "row identifier" for create/delete? Why cannot we use, say, `row_id` or `pk`?

skyzh:
We don't have pk. Row handler is like a row id, but row handler contains file id + row id, which means that after compaction, a record's row id could change.

yinfredyue:
> We don't have pk.

Does that mean RisingLight does not store pks? I am pretty sure you can have pk when creating tables.

```
pub struct BoundCreateTable {
    pub database_id: DatabaseId,
    pub schema_id: SchemaId,
    pub table_name: String,
    pub columns: Vec<ColumnCatalog>,
    pub ordered_pk_ids: Vec<ColumnId>,
}
```

> Row handler is like a row id, but row handler contains file id + row id, which means that after compaction, a record's row id could change.

So row_handler = file id + row id. Does the row_id equal to a row's index in the entire table?

skyzh:
No. It's the offset in the RowSet.

PR #698, opened 2022-09-30
bug: fix TypeMismatch panic at compaction

### LINKED ISSUE #696: bug: compaction panic
When working on #645, I found this bug where the compactor panics after two insertions.

image

I have looked into this. Will provide a more detailed description of the bug, and submit a PR to fix it.

yinfredyue:
Trying to fix #696

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,9 +30,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.19"
+version = "0.7.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b4f55bd91a0978cbfd91c457a164bab8b4001c833b7f323132c0a4e1922dd44e"
+checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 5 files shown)


skyzh:
`From<&array::ArrayImpl>` is implemented on concrete types (I32Array, Utf8Array, etc.) instead of the generic `Array` trait. Therefore, you have to use macros to expand them instead of calling it on traits.

yinfredyue:
PTAL @skyzh

yinfredyue:
SQL execution before this fix:

image

SQL execution after the fix: 

image

src/array/shuffle_ext.rs
```diff
@@ -141,11 +133,10 @@ macro_rules! impl_array_impl_shuffle_ext {
                 match self {
                     $(
                         Self::$Abc(builder) => {
-                            let typed_arrays = arrays
-                                .iter()
-                                .map(|x| x.try_into().unwrap())
-                                .collect::<SmallVec<[_; 8]>>();
-                            builder.pick_from_multiple(&typed_arrays, logical_rows);
+                            for (idx, row) in logical_rows {
+                                let arr: & $AbcArray = (&arrays[*idx]).try_into().unwrap();
```
skyzh, line 137: This is inefficient because we need to branching inside a loop to do the cast every time. Can we move arr outside?

PR #700, opened 2022-10-02
feat: make binary operations auto-vectorized

lokax:
There are null values in the last 64 pieces of every 192 pieces of data.
```rust
array mul binary op/65536                                                                            
                        time:   [590.75 µs 592.65 µs 595.27 µs]

array mul function (standard)/65536                                                                            
                        time:   [73.460 µs 73.719 µs 74.018 µs]

array mul function (standard + never inline )/65536                                                                            
                        time:   [177.76 µs 178.30 µs 178.88 µs]

array mul function (lazy select)/65536                                                                             
                        time:   [25.540 µs 25.698 µs 25.869 µs]

array mul function (lazy select + never inline)/65536                                                                            
                        time:   [109.85 µs 110.21 µs 110.70 µs]

array mul simd/65536    time:   [108.47 µs 108.87 µs 109.35 µs]

```


skyzh (approved):

skyzh:
There are a few commits without sign offs. Could you please fix it? Thanks!

Issue #679, opened 2022-08-04
storage: should_finish function for dict encoding is always true 


lokax:
https://github.com/risinglightdb/risinglight/blob/c859dffc379f8ce78737d60f0e95624766cd0a2b/src/storage/secondary/block/dict_block_builder.rs#L39

If block_builder has no data, the block_builder.estimated_size() will return 0. the target size of rle builder is 0. rle_builder.should_finished()  is always true.

```rust
fn should_finish(&self, next_item: &Option<&A::Item>) -> bool {
        self.data_builder.should_finish(next_item)
            || self
                .rle_builder
                .should_finish(&Some(&(DICT_NULL_VALUE_KEY)))
    }
```

unconsolable:
@skyzh This issue seems can be closed, ref https://github.com/risinglightdb/risinglight/pull/701#issuecomment-1264897994

PR #701, opened 2022-10-03
feat(storage): support dict encoding in primitive column builder

Kikkon:
🤣 #670 has been completed in issue #699, so we can support dict-encoding in primitive column builder.


Kikkon:
and fixed issue #679  by the way

PR #702, opened 2022-10-05
feat: introduce next-gen planner & optimizer

wangrunji0408:
This PR introduces a brand new optimizer based on [egg](https://egraphs-good.github.io), a powerful crate to build [cascades-style](https://zhuanlan.zhihu.com/p/73545345) optimizers in Rust.

With egg, we can define a complex rewrite rule using pattern matching in a few lines:
```rust
// predefine join node: (join type condition left right)
rewrite!("join-reorder";
    "(join ?type ?cond2 (join ?type ?cond1 ?left ?mid) ?right)" =>
    "(join ?type ?cond1 ?left (join ?type ?cond2 ?mid ?right))"
    if columns_is_disjoint("?cond2", "?left")
    // ^ a custom function to test whether the column set of two nodes are disjoint
),
```

As a result, the new optimizer is **super small but complete**.

In about 600 lines of code (+ 200 lines for test), it implements the following optimizations:
- expression simplification
- constant folding
- predicate pushdown
- nested-loop-join -> hash-join
- join reordering
- column pruning

and resolves two annoying problems in the old planner:
- extracting aggregations from select list
  ```sql
  SELECT sum(a) + b FROM t GROUP BY b;
  => 
  SELECT s + b FROM          # projection
    (SELECT b, sum(a) AS s   # aggregation
     FROM t GROUP BY b);
  ```
- converting column references to physical indices (InputRef)
  and maintaining them during optimization
  ```sql
  =>
  SELECT $1 + $0 FROM 
    (SELECT b, sum(a) FROM t GROUP BY b);
  ```

As a comparison, the current planner + optimizer has nearly 4100 lines of code (+ 600 lines for test) with a similar feature set.

In my opinion, the egg framework is definitely a game changer for building database optimizers.
If it proves to work well in RisingLight, we can consider migrating it to RisingWave in the future.


skyzh:
Looks great to me! Will take a look soon.

src/planner/rules/agg.rs
```diff
@@ -0,0 +1,157 @@
...
+        (filter ?having
+        (projagg ?exprs ?groupby
+        (filter ?where
+        ?from
+        )))))"
+    ),
+    rw!("split-projagg";
+        "(projagg ?exprs ?groupby ?child)" =>
+        { ExtractAgg {
+            has_agg: pattern("(proj ?exprs (agg ?aggs ?groupby ?child))"),
```
skyzh, line 20: There could be agg in having clause. Maybe we will need more test cases to cover this.

src/catalog/mod.rs
```diff
@@ -85,6 +86,45 @@ impl ColumnRefId {
     }
 }
 
+impl std::fmt::Display for ColumnRefId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // TODO: now ignore database and schema
+        write!(f, "${}.{}", self.table_id, self.column_id)
```
skyzh, line 92: We should make ColumnRefId more generic (e.g. `Vec<String>`) so as to represent a column in subquery. e.g. `subquery.table1.x`. I'll do this probably when refactoring binder.

src/planner/rules/schema.rs
```diff
@@ -0,0 +1,104 @@
...
+use crate::types::ColumnIndex;
+
+/// Replaces all column references (`ColumnRefId`) with
+/// physical indices ([`ColumnIndex`]) to the given schema.
+///
+/// # Example
+/// - given schema:           `sum(v1), v2`
+/// - the expressions:        `v2 + 1, sum(v1) + v2`
+/// - should be rewritten to: `#1 + 1, #0 + #1`
+pub fn resolve_column_index(expr: RecExpr, schema: &[RecExpr]) -> RecExpr {
```
skyzh, line 15: This is exactly what I want to do for a long time. We should not have column index until the very last stage of planning.

skyzh (approved):
Generally LGTM, thanks for your great work! It looks like the planner has not been fully integrated with the current query processing code path. Is there a tracking issue / TODO list for this?

wangrunji0408:
> It looks like the planner has not been fully integrated with the current query processing code path. Is there a tracking issue / TODO list for this?

Not yet. I'll create a tracking issue soon.

A rough plan is to:
1. fork the current binder, convert the sqlparser AST into the new egg AST.
2. rewrite the type checking as an egg analysis. (optional)
3. construct executors from the egg AST.
4. remove the old binder, planner and optimizer

src/planner/rules/schema.rs
wangrunji0408: ++++1

src/catalog/mod.rs
wangrunji0408: I gave up taking subquery into account. It's a little complicated. 🤯

src/planner/rules/agg.rs
wangrunji0408: You are right. There is a bug here. 🥵

skyzh:
Sounds great. I'll start working on the new binder soon. Constructing executors from egg AST also doesn't look hard.

src/planner/rules/agg.rs
wangrunji0408: Fixed in b8edc65. PTAL!

skyzh:
ready to merge?

src/planner/rules/agg.rs
wangrunji0408: Still not correct. The HAVING expression should be extracted as well.
Thanks for your great article. 📖

fuyufjh:
Really blow my mind!

fuyufjh:
Well, to be honest, I still had some concerns about the capability of DSL for a productional database system.

https://twitter.com/fuyufjh/status/1509182041828790274?s=20&t=LBlWsQKaNTiU2QBXbDtGBg

wangrunji0408:
I think these kinds of DSL are suitable to describe the high-level structural trait of a rule. The rest detailed and complex logic is usually left to the native language. In fact, this combination is a common pattern in the Programming Language field (e.g. lex, yacc...). Because functional, declarative languages are good at language processing, while low-level procedural languages are more efficient. In this term, although Rust has pattern matching, we found it's not expressive enough for complex rules. So in my opinion, some kinds of meta programming are still needed, whatever it is macro, codegen or DSL.

src/planner/rules/agg.rs
wangrunji0408: fixed in #706

CAJan93:
Thank you very much for your work! Looks amazing!

PR #704, opened 2022-10-06
fix(storage): checksum BlockType in block header

### LINKED ISSUE #247: storage: checksum `BlockType` in block header

unconsolable:
What's changed:
- Split the encode of `BlockHeader` to checksum and non-checksum
- Take non-checksum part of header (currently BlockType) to checksum calculation
- Move `BlockHeader` to the footer of block
- Change checksum validation logic

close #247 

### DIFF
--- a/src/storage/secondary/column.rs
+++ b/src/storage/secondary/column.rs
@@ -37,8 +37,8 @@ use bytes::Bytes;
 pub use char_column_factory::*;
 use moka::future::Cache;
 
-use super::block::BLOCK_HEADER_CHECKSUM_SIZE;
-use super::{Block, BlockCacheKey, BlockHeader, ColumnIndex, BLOCK_HEADER_SIZE};
+use super::block::BLOCK_META_CHECKSUM_SIZE;
+use super::{Block, BlockCacheKey, BlockMeta, ColumnIndex, BLOCK_META_SIZE};
 use crate::array::Array;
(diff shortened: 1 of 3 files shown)


src/storage/secondary/column.rs
```diff
@@ -195,18 +196,17 @@ impl Column {
                 "block is smaller than header size",
             ));
         }
-        let mut header = &block[..BLOCK_HEADER_SIZE];
-        let block_data = &block[BLOCK_HEADER_SIZE..];
+        let mut header = &block[block.len() - BLOCK_HEADER_SIZE..];
```
skyzh, line 199: should BLOCK_HEADER_SIZE be changed to BLOCK_FOOTER_SIZE?

skyzh (approved):
Rest LGTM. If you don't want to resolve the comments for now, ping me for merging and I'll do the rest.

src/storage/secondary/column.rs
unconsolable: Shall the struct name `BlockHeader` be renamed as `BlockFooter` as well?

skyzh: Yes, +1 for this!

skyzh: Or BlockMeta, whatever you want.

PR #705, opened 2022-10-09
feat(binder): introduce egg binder for queries 

wangrunji0408:
This PR forks the current binder and integrates it with the egg language introduced by #702.
(only queries for now, other statements are left to future PR)

Now the new planner is partially connected to the pipeline: 
SQL ==parser=> Parser AST ==binder=> Egg AST ==optimizer=> Optimized Egg AST.

You can try it by first typing `\v2` and then running queries in the interactive shell:
```sql
$ cargo run
> \v2
switched to planner v2
in 0.000s
# TPC-H Q3
> select
    l_orderkey,
    sum(l_extendedprice * (1 - l_discount)) as revenue,
    o_orderdate,
    o_shippriority
from
    customer,
    orders,
    lineitem
where
    c_mktsegment = 'BUILDING'
    and c_custkey = o_custkey
    and l_orderkey = o_orderkey
    and o_orderdate < date '1995-03-15'
    and l_shipdate > date '1995-03-15'
group by
    l_orderkey,
    o_orderdate,
    o_shippriority
order by
    revenue desc,
    o_orderdate
limit 10;
bind result:
(topn
  10
  null
  (list (desc (sum (* $7.5 (- 1 $7.6)))) (asc $6.4))
  (select
    (list $7.0 (sum (* $7.5 (- 1 $7.6))) $6.4 $6.7)
    (join
      cross
      true
      (join
        cross
        true
        (scan (list $5.0 $5.1 $5.2 $5.3 $5.4 $5.5 $5.6 $5.7))
        (scan (list $6.0 $6.1 $6.2 $6.3 $6.4 $6.5 $6.6 $6.7 $6.8)))
      (scan (list $7.0 $7.1 $7.2 $7.3 $7.4 $7.5 $7.6 $7.7 $7.8 $7.9 $7.10 $7.11 $7.12 $7.13 $7.14 $7.15)))
    (and
      (and
        (and (and (= $5.6 'BUILDING') (= $5.0 $6.1)) (= $7.0 $6.0))
        (< $6.4 1995-03-15))
      (> $7.10 1995-03-15))
    (list $7.0 $6.4 $6.7)
    true))
optimize result:
(topn
  10
  null
  (list (desc (sum (* $7.5 (- 1 $7.6)))) (asc $6.4))
  (proj
    (list $7.0 (sum (* $7.5 (- 1 $7.6))) $6.4 $6.7)
    (agg
      (list (sum (* $7.5 (- 1 $7.6))))
      (list $7.0 $6.4 $6.7)
      (join
        cross
        (= $7.0 $6.0)
        (filter
          (> 1995-03-15 $6.4)
          (join
            cross
            (and (= $6.1 $5.0) (= 'BUILDING' $5.6))
            (scan (list $5.0 $5.6))
            (scan (list $6.0 $6.1 $6.4 $6.7))))
        (filter
          (> $7.10 1995-03-15)
          (scan (list $7.0 $7.5 $7.6 $7.10)))))))
```

As you can see, the rewrite rules work but the output is not optimal, because we don't have a proper cost function. This will be the focus of the next step.

### DIFF
--- a/src/binder_v2/mod.rs
+++ b/src/binder_v2/mod.rs
@@ -9,7 +9,7 @@ use itertools::Itertools;
 
 use crate::catalog::{RootCatalog, TableRefId, DEFAULT_DATABASE_NAME, DEFAULT_SCHEMA_NAME};
 use crate::parser::*;
-use crate::planner::{Expr as Node, RecExpr};
+use crate::planner::{Expr as Node, ExprAnalysis, RecExpr};
 use crate::types::{DataTypeKind, DataValue};
 
 mod expr;
@@ -62,7 +62,7 @@ pub enum BindError {
 /// The binder resolves all expressions referring to schema objects such as
 /// tables or views with their column names and types.
 pub struct Binder {
-    egraph: egg::EGraph<Node, ()>,
+    egraph: egg::EGraph<Node, ExprAnalysis>,
     catalog: Arc<RootCatalog>,
     contexts: Vec<Context>,
 }
--- a/src/binder_v2/mod.rs
+++ b/src/binder_v2/mod.rs
@@ -1,0 +1,168 @@
+// Copyright 2022 RisingLight Project Authors. Licensed under Apache-2.0.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::vec::Vec;
+
+use egg::Id;
+use itertools::Itertools;
+
+use crate::catalog::{RootCatalog, TableRefId, DEFAULT_DATABASE_NAME, DEFAULT_SCHEMA_NAME};
+use crate::parser::*;
+use crate::planner::{Expr as Node, RecExpr};
+use crate::types::{DataTypeKind, DataValue};
+
+mod expr;
+mod select;
+mod table;
+
+pub use self::expr::*;
+pub use self::select::*;
+pub use self::table::*;
+
+pub type Result<T = Id> = std::result::Result<T, BindError>;
+
+/// The error type of bind operations.
+#[derive(thiserror::Error, Debug, PartialEq, Eq)]
+pub enum BindError {
+    #[error("invalid database {0}")]
+    InvalidDatabase(String),
+    #[error("invalid schema {0}")]
+    InvalidSchema(String),
+    #[error("invalid table {0}")]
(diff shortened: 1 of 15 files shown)


src/binder_v2/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    egraph: egg::EGraph<Node, ExprAnalysis>,
+    catalog: Arc<RootCatalog>,
+    contexts: Vec<Context>,
+}
+
+/// The context of binder execution.
+#[derive(Debug, Default)]
+struct Context {
+    /// Mapping table name to its ID.
+    tables: HashMap<String, TableRefId>,
```
skyzh, line 74: I'll prefer make Binder stateless in the future. Will do the refactor later.

skyzh (approved):
LGTM, good work!

skyzh:
... by the way, is it possible to implement a custom formatter for query plans? 🤣 Feel like it's hard to pair `(` and `)`.

skyzh:
Also it seems that we already resolve all columns to its offset in binder. But the schema should be unknown in the binding stage... Sounds like another thing that I'll do the refactor.

src/binder_v2/mod.rs
wangrunji0408: Do you mean to make the context an analysis data associated with each node?
There would be more overhead to maintain the context for each node, but significantly reduce the possibility to get things wrong. So I would +1 for this proposal.

skyzh: I think we are on the same page. e.g. if we want to resolve a column name, we look into the BoundTableRef (or the context bound to the table ref) instead of looking at the global context.

wangrunji0408:
> ... by the way, is it possible to implement a custom formatter for query plans? 🤣 Feel like it's hard to pair ( and ).

Surely we will define our explain formatter as before. The current format is just for a quick preview. 🤣

> Also it seems that we already resolve all columns to its offset in binder. But the schema should be unknown in the binding stage...

Analyzing schema in the binder is for replacing wildcard `*` with a list of expressions that the sub-plan outputs. I doubt the implementation in the old binder is not correct in the sub-query case. 😇

PR #709, opened 2022-10-13
feat(storage): generic nullable block encoding

### LINKED ISSUE #533: storage: generic nullable block encoding
Currently, the nullable encoding is done by using a bitmap, and is only implemented for primitive types.

https://github.com/risinglightdb/risinglight/blob/7b5bc7a1a258afd64a483990ca95f4c799482791/src/storage/secondary/block/primitive_nullable_block_builder.rs#L18

Indeed, such logic can be done in a more generic way. We can wrap any BlockIterator to become a nullable encoding.

```
pub struct NullableBlockBuilder<B: BlockBuilder> {
  inner: B,
  bitmap: BitVec
}
```

To achieve this, we can:

- [ ] separate nullable encoding from primitive nullable encoding.
- [ ] use the generic nullable encoder for char and blob.
- [ ] to avoid branch prediction for `Option::unwrap`, we can make a new pair of traits: `NonNullableBlockBuilder` and `Iterator`, which accepts `&A::Item` instead of `Option<&A::Item>`.

unconsolable:
What's changed:
- Introduce generic nullable block encoding
- Support nullable encoding on blob, primitive, char
- Add new pair of traits `NonNullableBlockBuilder` & `NonNullableBlockIterator`

close #533

### DIFF
--- a/src/storage/secondary/block/blob_block_builder.rs
+++ b/src/storage/secondary/block/blob_block_builder.rs
@@ -42,7 +42,7 @@ impl<T: BlobEncode + ?Sized> NonNullableBlockBuilder<T::ArrayType> for PlainBlob
     }
 
     fn append_default(&mut self) {
-        self.data.extend("".to_byte_slice());
+        // don't extend `self.data` since empty value is default
         self.offsets.push(self.data.len() as u32);
     }
 
--- a/src/storage/secondary/block/blob_block_builder.rs
+++ b/src/storage/secondary/block/blob_block_builder.rs
@@ -2,11 +2,13 @@
 
 use std::marker::PhantomData;
 
+use bitvec::prelude::{BitVec, Lsb0};
 use bytes::BufMut;
 use risinglight_proto::rowset::BlockStatistics;
 
 use super::super::statistics::StatisticsBuilder;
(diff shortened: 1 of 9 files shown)


src/storage/secondary/block/blob_block_builder.rs
```diff
@@ -33,13 +35,45 @@ impl<T: BlobEncode + ?Sized> PlainBlobBlockBuilder<T> {
...
-    fn append(&mut self, item: Option<&T>) {
-        let item = item.expect("nullable item found in non-nullable block builder");
+impl<T: BlobEncode + ?Sized> NonNullableBlockBuilder<T::ArrayType> for PlainBlobBlockBuilder<T> {
+    fn append_value(&mut self, item: &<T::ArrayType as Array>::Item) {
         self.data.extend(item.to_byte_slice());
         self.offsets.push(self.data.len() as u32);
     }
 
+    fn append_default(&mut self) {
+        self.data.extend("".to_byte_slice());
```
skyzh, line 45: This step is not necessary. It is equivalent to no-op.

skyzh (approved):
Rest LGTM, good work!

Issue #707, opened 2022-10-13
binder v2 for statements


wangrunji0408:
#705 introduced a new binder for egg expressions. But it only covered `SELECT` queries. We need to do the same for the rest statements.

skyzh:
@eliasyaoyc do you have any overview of how it would goes in mind? If not, I think it's better for Runji and me to move forward on this issue.

PR #711, opened 2022-10-14
feat(binder): egg binder for drop statement

eliasyaoyc:
#707

### DIFF
--- a/src/binder_v2/drop.rs
+++ b/src/binder_v2/drop.rs
@@ -1,0 +1,137 @@
+use std::result::Result as RawResult;
+use std::str::FromStr;
+
+use serde::{Deserialize, Serialize};
+
+use super::*;
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Copy, Clone, Serialize, Deserialize)]
+pub struct BindDrop {
+    pub object: Object,
+    pub if_exists: bool,
+    pub cascade: bool,
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Copy, Clone, Serialize, Deserialize)]
+pub enum Object {
+    Table(TableRefId),
+}
+
+impl std::fmt::Display for BindDrop {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "object: {}, exists: {}, cascade: {}",
+            self.object, self.if_exists, self.cascade,
+        )
+    }
+}
+
+impl std::fmt::Display for Object {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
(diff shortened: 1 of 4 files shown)


src/binder_v2/drop.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub(super) fn bind_drop(
+        &mut self,
+        object_type: ObjectType,
+        if_exists: bool,
+        names: Vec<ObjectName>,
+        cascade: bool,
+    ) -> Result {
+        self.push_context();
+        let ret = self.bind_drop_internal(object_type, if_exists, names, cascade);
+        self.pop_context();
```
wangrunji0408, line 84: Drop statement won't be nested, so there's no need to push another context.

src/binder_v2/drop.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[error("invalid number: {0}")]
+    InvalidNum(#[from] std::num::ParseIntError),
+    #[error("invalid bool: {0}")]
+    InvalidBool(#[from] std::str::ParseBoolError),
+}
+
+impl FromStr for BindDrop {
+    type Err = ParseDropError;
+
+    fn from_str(s: &str) -> RawResult<Self, Self::Err> {
```
wangrunji0408, line 54: `FromStr` is used for parsing an egg expression from a LISP-style S-expression. It's hard to design a format for such a structure and maybe we will never construct one from the string. So I suggest just returning an error in this function like:
```rust
impl FromStr for BindDrop {
    type Err = ();
    fn from_str(s: &str) -> Result<Self, Self::Err> { Err(()) }
}
```

wangrunji0408 (commented):
Rest LGTM. Thanks!

wangrunji0408 (approved):

skyzh:
Indeed I'm wondering why we're commenting out the original drop and add a new BoundDrop between the expressions. Please hold this PR a little while and I will refactor tonight.

wangrunji0408:
> Indeed I'm wondering why we're commenting out the original drop and add a new BoundDrop between the expressions. Please hold this PR a little while and I will refactor tonight.

I thought both ways are acceptable. Given that the drop statement will unlikely be rewritten by rules, representing it as a single `BoundDrop` value can avoid parsing it back from the `Expr` in the executor. 🤔

skyzh (approved):
LGTM. Some futures plans in my mind:

* Why insert / create / drop are part of `pub enum Expr`? 🤣 Seems a little bit weird but I'm thinking if it is possible to make it more clear.
* I believe we only need to support Drop(Id) in RisingLight. We can remove cascade / remove_if_exist later.

wangrunji0408:
> Why insert / create / drop are part of pub enum Expr? 🤣 Seems a little bit weird but I'm thinking if it is possible to make it more clear.

Yeah, putting them all into the `Expr` is just for making a single AST representation in the planner, although these statements won't benefit much from the egg's features since there is no need to rewrite or analyze them.

PR #716, opened 2022-10-17
refactor(fetch size): add fast return and compare ROWSET MAX OUTPUT

eliasyaoyc:
Fast return: when all columns size is `0`, only has tow case:
1. index of current block is no data can fetch (use `ROWSET_MAX_OUTPUT`).
2. all columns is finished (return directly).
(old code,  if all columns is finished then return 0 and io will occur.)

Signed-off-by: Elias.Yao <siran0611@gmail.com>


skyzh:
Will come back and take a look soon. Sorry for the delay.

skyzh (approved):

PR #719, opened 2022-10-27
feat(binder): new binder for copy, delete and create_table

eliasyaoyc:
Signed-off-by: Elias.Yao <siran0611@gmail.com>


eliasyaoyc:
@wangrunji0408  cc

PR #721, opened 2022-10-31
fix(type): add NULL type and fix type analysis

wangrunji0408:
This PR adds a NULL variant to the `DataTypeKind` enum, as the type of NULL value. And fixes the remaining failed tests in type analysis.

### DIFF
--- a/src/array/mod.rs
+++ b/src/array/mod.rs
@@ -339,6 +339,7 @@ macro_rules! impl_array_builder {
             pub fn with_capacity(capacity: usize, ty: &DataType) -> Self {
                 use DataTypeKind::*;
                 match ty.kind() {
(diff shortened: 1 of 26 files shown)


skyzh:
In general NULL should be part of the type itself (e.g., int null, varchar null) instead of a separate data type kind? Though I'm okay with this change.

wangrunji0408:
> In general NULL should be part of the type itself (e.g., int null, varchar null) instead of a separate data type kind?

Do you mean "nullable"? (e.g. int nullable, varchar not nullable)

The null type in this PR means an "unknown" type for a single NULL value. Previously we represent this by a `None` in `Option<DataType>`, and now it will be `DataType { kind: Null, nullable: true }`.

skyzh:
Oh that seems fine!

skyzh (approved):
Good work!

PR #724, opened 2022-11-02
feat(storage): support NULL on blob and char columns

### LINKED ISSUE #718: support `NULL` on blob and char columns
Since https://github.com/risinglightdb/risinglight/issues/712 is merged, may `NULL` on blob and char columns can be supported.

Currently, `NULL` on these columns will become default value.

```
> create table test (e varchar(10));
created
in 0.031s
> insert into test values (NULL);
1 rows inserted
in 0.035s
> select * from test;
+---+
| e |
+---+
|   |
+---+
in 0.003s
```

unconsolable:

close #718

- Add nullable type enum in proto
- Add new fn `is_empty` in trait `NonNullableBlockBuilder`, to fix `should_finish` is true while inner builder is empty.
- Add enums in `column_builder` & `column_factory`, and support these enums in `match` (Few marcos are used to eliminate common code.)


skyzh (approved):
Thanks!

PR #728, opened 2022-11-09
feat(planner_v2): pass all sqllogictests

wangrunji0408:
This PR fixes various problems in the new planner and makes it pass all sqllogictests.

Detailed changes:

- Introduce the `NullArray`. The only valid element in this array is NULL. It can be casted into any other type of array.
- Allow `DataChunk` to have 0 column but non-zero cardinality. This will appear when all columns are pruned in a table. e.g `select count(*) from t`, `select t.a from t, s`.
- Add table_id to `Scan` node. As mentioned above, even when all columns are pruned, the scan node should emit chunks with correct cardinality. So at least it needs to know which table to scan.
  - @skyzh The disk storage doesn't support scanning a table with no column and emitting empty chunks with cardinality only. Is it possible to implement it? Now, as a workaround, the TableScan executor will append a row-handler column for the storage and discard it from the chunk later. 
- Floating-point constants are parsed to decimal instead of f64, in order to avoid precision issues.
- Support DELETE statement. We add an internal column `_row_id_` with column_id = u32::MAX to represent the row handler. This will be used in planning DELETE statement.
- Move agg extraction from rules to binder, for better error handling. Then the `Select` and `Distinct` node can be removed.
- Support more operations with different types in the evaluator. Fix the state of `count`, `first` and `last` agg.
- Introduce the `Empty` node to represent an optimized plan with no output.
- Enable sqllogictest for both v1 and v2 query engine.

After this PR, the new query engine is fully ready to use!


skyzh:
> The disk storage doesn't support scanning a table with no column and emitting empty chunks with cardinality only. Is it possible to implement it? Now, as a workaround, the TableScan executor will append a row-handler column for the storage and discard it from the chunk later.

The expected way to do that is to return a chunk with only row handle column. There's no way of producing chunk of cardinality very efficiently right now in RisingLight.

skyzh (approved):
Great work!

PR #730, opened 2022-11-10
feat(rt): support tokio-console

wangrunji0408:
This PR spawns each executor as a tokio task and adds support for debugging with [tokio-console](https://tokio.rs/blog/2021-12-announcing-tokio-console).

Running with `cargo r -- --tokio-console` and opening `tokio-console`, you will have a clear view of all executors and background tasks:

risinglight-tokio-console

This will be useful in performance tuning.

This PR also upgrades dependencies and fixes issues caused by sqlparser API change.


xxchan (commented):
What's the rationale of spawning a tokio task for each executor? 🤔

src/v1/optimizer/logical_plan_rewriter/arith_expr_simplification.rs
```diff
@@ -105,38 +105,3 @@ impl PlanRewriter for ArithExprSimplificationRule {
...
-    use std::fs;
-
-    use manifest_dir_macros;
-    use serde::Serialize;
-
-    use crate::Database;
-
-    #[tokio::main]
-    #[test]
-    async fn test_expr_simplification() {
```
xxchan, line 120: Why is it removed?

wangrunji0408: I thought it had been replaced by sqlplannertest?

wangrunji0408:
> What's the rationale of spawning a tokio task for each executor? 🤔

To make executors visible in tokio-console and allow parallel execution on a multi-thread runtime.

xxchan:
Are there any disadvantages to have too many tasks? 🤔 (BTW how does risingwave do it?

xxchan (approved):

skyzh:
> What's the rationale of spawning a tokio task for each executor? 🤔

In the future when we implement intra-operator parallelism we should spawn a tokio task for each stage instead of for each executor.

skyzh (approved):

skyzh:
> Are there any disadvantages to spawn too many tasks? 🤔 (BTW how does risingwave do it?

Both streaming and batch engine spawns a task for a fragment.

wangrunji0408:
> Are there any disadvantages to spawn too many tasks? 🤔 

If cancellation is not handled properly, it can leave orphan tasks and leak resources. In addition, too many tasks may cause problems in scheduling, like high latency. But I think in our use case (one task per executor), the number is far less than the limit. So it should be fine. 🤔

PR #732, opened 2022-11-14
chore: bump toolchain and apply async-fn-in-trait

wangrunji0408:
This PR applies the latest feature async-fn-in-trait (AFIT) to the traits and their impls in storage. But AFIT is still incomplete and can not properly infer Send. So for the trait definations, we use return-position-impl-trait-in-traits (RPITIT) as a workaround.


skyzh (approved):
LGTM. Though I generally discourage using incomplete features, the change seems easily revertible if something goes wrong and I'd happy to get it merged.

PR #754, opened 2022-12-29
feat(optimizer): upgrade column pruning to projection pushdown

wangrunji0408:
Signed-off-by: Runji Wang <wangrunji0408@163.com>

This PR upgrades **column pruning** to **projection pushdown**. Additional projections will be inserted to the plan tree, so that columns can be pruned in time after their last use.

The implementation of projection pushdown is similar to the original column pruning. Instead of generating and pushdown the `prune` node, now we directly generate the projection node with the used columns as projection. Then identical projections will be removed by the `identical-proj` rule.

Here is the execution plan of TPC-H Q10 before and after enabling projection pushdown:
```diff
 Projection: [c_custkey, c_name, sum((l_extendedprice * (1 - l_discount))), c_acctbal, n_name, c_address, c_phone, c_comment]
   TopN: limit=20, offset=0, orderby=[sum((l_extendedprice * (1 - l_discount))) desc]
     Aggregate: [sum((l_extendedprice * (1 - l_discount)))], groupby=[c_custkey, c_name, c_acctbal, c_phone, n_name, c_address, c_comment]
+      Projection: [n_name, c_custkey, c_name, c_address, c_phone, c_acctbal, c_comment, l_extendedprice, l_discount]
         HashJoin: inner, on=([c_nationkey] = [n_nationkey])
+          Projection: [c_custkey, c_name, c_address, c_nationkey, c_phone, c_acctbal, c_comment, l_extendedprice, l_discount]
+            HashJoin: inner, on=([o_orderkey] = [l_orderkey])
+              Projection: [c_custkey, c_name, c_address, c_nationkey, c_phone, c_acctbal, c_comment, o_orderkey]
                 HashJoin: inner, on=([c_custkey] = [o_custkey])
                   Scan: customer[c_custkey, c_name, c_address, c_nationkey, c_phone, c_acctbal, c_comment]
-                  HashJoin: inner, on=([o_orderkey] = [l_orderkey])
+                  Projection: [o_orderkey, o_custkey]
                     Filter: ((o_orderdate >= 1993-10-01) and (1994-01-01 > o_orderdate))
                       Scan: orders[o_orderkey, o_custkey, o_orderdate]
+              Projection: [l_orderkey, l_extendedprice, l_discount]
                 Filter: (l_returnflag = 'R')
                   Scan: lineitem[l_orderkey, l_extendedprice, l_discount, l_returnflag]
           Scan: nation[n_nationkey, n_name]
```

As unused columns are removed before HashJoin, the join executor runs faster because there're fewer columns to build. The execution time of this query drops from 0.5s to 0.4s on my computer.

## Limitations

Projection nodes generated between join nodes prevent the join reordering rules. Now join reordering is temporarily unavailable. 😇

### DIFF
--- a/tests/planner_test/tpch.planner.sql
+++ b/tests/planner_test/tpch.planner.sql
@@ -112,11 +112,12 @@ order by
     l_linestatus;
 
 /*
-Projection: [l_returnflag, l_linestatus, sum(l_quantity), sum(l_extendedprice), sum((l_extendedprice * (1 - l_discount))), sum(((1 - l_discount) * (l_extendedprice + (l_tax * l_extendedprice)))), (sum(l_quantity) / count(l_quantity)), (sum(l_extendedprice) / count(l_extendedprice)), (sum(l_discount) / count(l_discount)), rowcount] (cost=33878.984)
-  Order: [l_returnflag asc, l_linestatus asc] (cost=27798.984)
-    Aggregate: [sum(l_quantity), sum(l_extendedprice), sum((l_extendedprice * (1 - l_discount))), sum(((1 - l_discount) * (l_extendedprice + (l_tax * l_extendedprice)))), count(l_quantity), count(l_extendedprice), sum(l_discount), count(l_discount), rowcount], groupby=[l_returnflag, l_linestatus] (cost=19940)
-      Filter: (1998-09-21 >= l_shipdate) (cost=12900)
-        Scan: lineitem[l_quantity, l_extendedprice, l_discount, l_tax, l_returnflag, l_linestatus, l_shipdate] (cost=7000)
+Projection: [l_returnflag, l_linestatus, sum(l_quantity), sum(l_extendedprice), sum((l_extendedprice * (1 - l_discount))), sum(((l_tax + 1) * (l_extendedprice * (1 - l_discount)))), (sum(l_quantity) / count(l_quantity)), (sum(l_extendedprice) / count(l_extendedprice)), (sum(l_discount) / count(l_discount)), rowcount] (cost=39238.984)
+  Order: [l_returnflag asc, l_linestatus asc] (cost=33158.984)
+    Aggregate: [sum(l_quantity), sum(l_extendedprice), sum((l_extendedprice * (1 - l_discount))), sum(((l_tax + 1) * (l_extendedprice * (1 - l_discount)))), count(l_quantity), count(l_extendedprice), sum(l_discount), count(l_discount), rowcount], groupby=[l_returnflag, l_linestatus] (cost=25300)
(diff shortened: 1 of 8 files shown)


tests/planner_test/tpch.planner.sql
```diff
@@ -215,10 +227,11 @@ where
     and l_quantity < 24;
 
 /*
-Projection: [sum((l_discount * l_extendedprice))] (cost=7409.928)
-  Aggregate: [sum((l_discount * l_extendedprice))], groupby=[] (cost=7408.328)
-    Filter: ((l_shipdate >= 1994-01-01) and ((1995-01-01 > l_shipdate) and ((0.09 >= l_discount) and ((24 > l_quantity) and (l_discount >= 0.07))))) (cost=7210.72)
-      Scan: lineitem[l_quantity, l_extendedprice, l_discount, l_shipdate] (cost=4000)
+Projection: [sum((l_discount * l_extendedprice))] (cost=8163.5923)
+  Aggregate: [sum((l_discount * l_extendedprice))], groupby=[] (cost=8161.992)
+    Projection: [l_extendedprice, l_discount] (cost=7964.3843)
```
skyzh, line 232: From the execution point of view, I don't expect the new plan to be faster given this project is simply extracting two columns out of the data chunk.

skyzh (approved):
LGTM, good work! I wonder if this column pruning only works for HashJoin / TopN given they will need to reorder the source data in some ways. For the rest of the projections, we can remove it in the future. e.g., projection + aggregation (without distinct) should have no difference before / after column pruning.

tests/planner_test/tpch.planner.sql
wangrunji0408: Sure. The projection here is unnecessary. (But also won't be a big impact since it's a simple extraction.) It should be eliminated when the cost function gets smart enough to find it unnecessary. 😇

wangrunji0408:
Just found another problem. It will go wrong when there is an aggregation in the HAVING or GROUP BY clause.

```sql
SELECT b FROM t(a, b, c, d) GROUP BY b HAVING sum(a) > 1
```
```diff
    (proj (list b)
        (filter (> (sum a) 1)
+       (proj (list a b)
                    ^ should be (sum a)
            (agg (list (sum a)) (list b)
-               (scan t (list a b c d))
+               (scan t (list a b))
    ))))
```

Because the basic unit above the agg node should be `(sum a)` instead of `a`.
We should assign an alias to each aggregation result as we will do in subquery. cc @MingjiHan99

wangrunji0408:
> I wonder if this column pruning only works for HashJoin / TopN given they will need to reorder the source data in some ways. 

I think so.

> For the rest of the projections, we can remove it in the future. e.g., projection + aggregation (without distinct) should have no difference before / after column pruning.

I think it's also fine to just leave them here. The overhead should be negligible.

PR #755, opened 2022-12-30
feat(executor): bring back internal table scan executor to v2

### LINKED ISSUE #743: executor: bring back internal table scan executor to v2
```
select * from pg_catalog.contributors
```

doesn't work now, we may need to support it in future versions.

unconsolable:
close #743 

What's changed:
- Check whether the table is internal table in Binder
- Add `Internal` enum in planner `Node`
- Add `InternalTableExecutor` in executor v2
- Update contributors table
- Test internal table in integration test


src/planner/mod.rs
```diff
@@ -77,6 +77,7 @@ define_language! {
 
         // plans
         "scan" = Scan([Id; 2]),                 // (scan table [column..])
+        "internal" = Internal([Id; 2]),         // (internal table [column..])
```
wangrunji0408, line 80: I prefer not to introduce a new plan node for internal tables, otherwise those optimization rules for scan node can not be applied to internal table. Basically it is the same as `scan`, I think we could reuse the `scan` node until building executors.

wangrunji0408 (commented):
Generally LGTM except for the following issue. Thanks for your work!

src/planner/mod.rs
unconsolable: Got it. I wonder another question: is it acceptable to change the content of `contributors` to author name? I think that if it is acceptable, we can update the `contributors` contents in `build.rs` offline, otherwise `tokio` and `reqwest` may needed to call GitHub RESTful API, and `build-dependencies` may be much larger.

skyzh: I would prefer updating it manually. Otherwise build.rs might be slow and affect the normal build process.

skyzh: @wangrunji0408 but internal table doesn’t support filter push down like the normal storage engine? I guess it’s still a good idea to do it separately… not sure

skyzh: Also internal table doesn’t need to support columns pruning, in my opinion.

wangrunji0408 (approved):
LGTM! Please help resolve the conflicts. thx

src/planner/mod.rs
wangrunji0408: Sorry I missed this discussion. It seems reasonable to have the internal node. 🆒

PR #757, opened 2023-01-03
feat(api):support Postgres wire protocol

SkyFan2002:
This PR adds simple support for Postgres wire protocol.

For example:
First run risinglight with command line argument `--server`:
```bash
cargo run -- --server --host localhost --port 1234
```
Then connect it with psql:
```bash
psql -h localhost -p 1234 
```
And run a few sql statements:
```bash
vpt=> insert into t03 values(1,'a');
OK
vpt=> insert into t03 values(2,'b');
OK
vpt=> insert into t03 values(3,'c');
OK
vpt=> select * from t03;
++ | ++
----+----
1  | a
2  | b
3  | c
(3 rows)
vpt=> select id from t03;
++
----
 1
 2
 3
(3 rows)
vpt=> select a from t03;
ERROR:  bind error: invalid column a  
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,6 +268,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "block-buffer"
+version = "0.10.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


MingjiHan99:
LGTM, good work!

MingjiHan99 (approved):

PR #760, opened 2023-01-07
feat(storage): compact rowset & DV manifest in bootstrap

### LINKED ISSUE #739: storage: bootstrap manifest replay
When bootstrapping secondary storage, the manifest log is replayed to restore the internal state of `VersionManager`. This is done via the `commit_changes` call.

https://github.com/risinglightdb/risinglight/blob/365f2858099d856a6831eede1e75db1a5018a9b6/src/storage/secondary/storage.rs#L164

The problem with this is that any effective `AddRowset`/`AddDV` are written to the manifest file again at bootstrap. This means that, if the database keeps crashing and coming back, the manifest will keep growing (due to the bootstrap), even if no write happens.

Under the textbook WAL setting, redo&undo phase append more logs (and those logs provide useful information). But I don't think those logs are necessary/useful for RisingLight, because it's using a much simpler logging scheme.

unconsolable:

close #739 

What's changed:
- Rewrite effective `AddRowset` & `AddDV` with `CreateTable` & `DropTable` to temp manifest, then rename the temp manifest and reopen.
- Split `commit_changes` in order to support different manifest.


skyzh (approved):
LGTM, good work! In the future we might persist next table id in the manifest, so that we don't need to do extra work for them.

PR #761, opened 2023-01-07
feat: support TPC-H Q9

wangrunji0408:
With subquery support in #756, we are able to challenge more TPC-H queries.
As an example, this PR adds support for TPC-H Q9.

Changes:

- support `like` operator
- support `extract(year from xx)` function
- update `sqlparser-rs` version to 0.30 (for deriving`Ord`)
- add TPC-H Q9 to test

However, the speed for this query is ridiculously slow. (104s vs DuckDB <1s)
We should investigate and optimize it in the future.


MingjiHan99 (approved):
LGTM!

skyzh (approved):
LGTM. Q9 contains a lot of joins, and I guess we need to see if the join order is not optimal.

wangrunji0408:
Any idea about the failing CI? It only said `The operation was canceled.` 🥵

skyzh:
I guess it’s OOM or something…

PR #762, opened 2023-01-07
executor: use const generics to optimize hash join 

MingjiHan99:
Generate hash join code in the compilation time to avoid testing join types during the execution for **every row**.

### DIFF
--- a/src/executor_v2/mod.rs
+++ b/src/executor_v2/mod.rs
@@ -86,15 +86,15 @@ pub enum JoinType {
 }
 
 macro_rules! generate_hash_join_operator {
-    ($self:item, $join_type:expr, $op:expr, $lkeys:expr, $rkeys:expr, $left:expr, $right:expr) => {
-        return HashJoinExecutor::<{ $join_type }> {
-            op: self.node($op).clone(),
-            left_keys: self.resolve_column_index($lkeys, $left),
-            right_keys: self.resolve_column_index($rkeys, $right),
-            left_types: self.plan_types(left).to_vec(),
-            right_types: self.plan_types(right).to_vec(),
+    ($self:ident, $join_type:expr, $op:expr, $lkeys:expr, $rkeys:expr, $left:expr, $right:expr) => {
+        HashJoinExecutor::<{ $join_type }> {
+            op: $self.node($op).clone(),
+            left_keys: $self.resolve_column_index($lkeys, $left),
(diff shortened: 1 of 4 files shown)


MingjiHan99:
It is weird that all planner tests work on my machine .....

src/executor_v2/mod.rs
```diff
@@ -244,16 +253,47 @@ impl<S: Storage> Builder<S> {
...
+            HashJoin([op, lkeys, rkeys, left, right]) => {
+                if matches!(self.node(op), Expr::Inner) {
+                    HashJoinExecutor::<{ JoinType::Inner }> {
+                        op: self.node(op).clone(),
+                        left_keys: self.resolve_column_index(lkeys, left),
+                        right_keys: self.resolve_column_index(rkeys, right),
+                        left_types: self.plan_types(left).to_vec(),
+                        right_types: self.plan_types(right).to_vec(),
+                    }
+                    .execute(self.build_id(left), self.build_id(right))
```
xxchan, line 265: Use macro to simplify it?

MingjiHan99: Sure

wangrunji0408 (approved):
LGTM!

src/executor_v2/mod.rs
```diff
@@ -244,16 +265,19 @@ impl<S: Storage> Builder<S> {
...
-            HashJoin([op, lkeys, rkeys, left, right]) => HashJoinExecutor {
-                op: self.node(op).clone(),
-                left_keys: self.resolve_column_index(lkeys, left),
-                right_keys: self.resolve_column_index(rkeys, right),
-                left_types: self.plan_types(left).to_vec(),
-                right_types: self.plan_types(right).to_vec(),
+            HashJoin([op, lkeys, rkeys, left, right]) => {
+                if matches!(self.node(op), Expr::Inner) {
+                    hash_join_operator!(self, JoinType::Inner, op, lkeys, rkeys, left, right)
+                } else if matches!(self.node(op), Expr::LeftOuter) {
```
xxchan, line 271: Use `match` instead of `if matches!`? (It's surprising that there's no clippy warning? 🤔)

BTW just a personal preference not a requirement: I would define the macro just here, and catch local variables directly without passing arguments. 🤪

xxchan (approved):

Issue #767, opened 2023-01-10
type: support TIMESTAMP


wangrunji0408:
Reference of adding a type: #766

SkyFan2002:
#769

PR #769, opened 2023-01-14
feat(type): support timestamp and timestamptz

SkyFan2002:


tests/sql/timestamp.sql
```diff
@@ -0,0 +1,49 @@
...
+values ('1991-01-14 04:05:06 +08:00');
+insert into timestamp_test
+values ('1991-01-15 04:05:06 AD +08:00');
+insert into timestamp_test
+values ('1991-01-16 04:05:06 BC +08:00');
+insert into timestamp_test
+values ('1991-01-17 04:05:06 +08:00 AD');
+insert into timestamp_test
+values ('1991-01-18 04:05:06 +08:00 BC');
+select *
```
unconsolable, line 21: IMHO, it might be better to use *.slt format, which can reflect query results. Here is E2E integration test guide https://github.com/risinglightdb/risinglight/blob/main/docs/05-e2e-tests.md

SkyFan2002: Thanks fot pointing out this.I am tring to solve it but meet an error as the ci shows.Do you have any idea about this?

tests/sql/timestamp.slt
```diff
@@ -0,0 +1,83 @@
+statement ok
```
unconsolable, line 1: Switch from CRLF to LF

tests/sql/timestamp.sql
unconsolable: close #767

unconsolable:
close #767

tests/sql/timestamp.sql
unconsolable: Also it might be better to add `close #767 ` to https://github.com/risinglightdb/risinglight/pull/769#issue-1533115570, which can help close a issue

wangrunji0408 (approved):
LGTM! Thanks!

PR #774, opened 2023-02-18
feat(catalog): remove the database level

wangrunji0408:
As a lightweight database, it is too complex to have a database-schema-table-column hierarchy in the catalog. We don't support creating databases or schemas until now and nobody has even complained about it. Besides, neither SQLite nor DuckDB has a database level. So I propose doing the same for RisingLight.

This PR removes the database level and does some refactoring in the catalog.


skyzh (approved):

skyzh:
Let me configure the merge queue...

PR #777, opened 2023-02-19
chore: bump deps version, patch nondeterministic in egg

skyzh:


skyzh:
It seems that the generated prettier result is not stable across platforms. Something wrong with UTF-8 with estimation?

xxchan:
cc @ice1000

xxchan (approved):

ice1000:
> It seems that the generated prettier result is not stable across platforms. Something wrong with UTF-8 width estimation?

I am using the standard `s.chars().count()`, is that stable across platforms?

xxchan (commented):
It's the order of the exprs that changed


image

xxchan:
It's not related to the prettier, run `make apply_planner_test` on 70acb0e will also cause a diff here 😅
image

ice1000:
Turns out using ordered btreemap was better than this lmao

ice1000:
By the way I have observed a problem with the pretty printing crate. I'll have to fix it. It should inline singleton lists

xxchan:
> Turns out using ordered btreemap was better than this lmao

ordered btreemap will prevent us from finding this thing 🤣

wangrunji0408:
I also noticed this problem before. I believe it's caused by the egg. Because the plan is generated by egg's runner and extracted using our cost function, where identical expressions (e.g. `a AND b`, `b AND a`) should have the same cost.
So any of them could be part of the best plan. But I have no idea why it is not stable across platforms. By the way, we also ran into this problem in deterministic testing of RisingWave. 🤪

wangrunji0408 (approved):

xxchan:
> But I have no idea why it is not stable across platforms. By the way, we also ran into this problem in deterministic testing of RisingWave

Guess there must be some non-determinism 🤪

skyzh:
Hmm, I'd prefer fixing non-determinism before merging this PR. I'll take a look later.

skyzh:
fixed by patching egg, probably we can submit a PR to upstream to add a `deterministic` feature?

xxchan:
successful！

ice1000:
> fixed by patching egg, probably we can submit a PR to upstream to add a `deterministic` feature?

Wtf how did you fix that?

skyzh:
by replacing hashmap -> index(btree)map

ice1000:
Lmao gotcha

mwillsey:
hi :wave: egg developer here. Very cool work! I'd be interested in fixing any sources of non-determinism. I see [your patch](https://github.com/egraphs-good/egg/compare/main...risinglightdb:egg:skyzh/determinstic), but as far as I can tell it should not matter! We were using a deterministic hash function (fxhash)... shouldn't that be enough?

skyzh:
Yeah, I'm also surprised why it happened. I guess it's because non-deterministic in resizing the hash table, but I didn't look into it... One of RisingLight developer developed https://github.com/madsim-rs/madsim, which can detect non-determinism in Rust.

skyzh:
I also guess it's a platform-dependent thing because it always yield the same value in a single platform...

mwillsey:
Ok, I have added this as the `deterministic` feature: https://github.com/egraphs-good/egg/commit/967a3db0c059fc17bbb2da8cc4d4c59eb5f093e0

mwillsey:
My guess is that somewhere in your code a hash value is being computed from a pointer (as a value), which would explain the non-det even with a deterministic hash function and the non-det across platforms (different allocators). Either way, indexmap (as used in the above patch) should fix it!

PR #771, opened 2023-01-25
feat(explain): reimplement explain using pretty-xmlish

ice1000:
Edit: here's an example from planner tests:

```
Projection { cost: 8163.5923, exprs: [ sum {  } ] }                                                                 
└── Aggregate { aggs: [ sum {  } ], cost: 8161.992, group_by: [] }                                                  
    └── Projection { cost: 7964.3843, exprs: [ l_extendedprice, l_discount ] }                                      
        └── Filter                                                                                                  
            ├── cond: and                                                                                           
            │   ├── lhs: >= { lhs: l_shipdate, rhs: 1994-01-01 }                                                    
            │   └── rhs: and                                                                                        
            │       ├── lhs: > { lhs: 1995-01-01, rhs: l_shipdate }                                                 
            │       └── rhs: and                                                                                    
            │           ├── lhs: >= { lhs: 0.09, rhs: l_discount }                                                  
            │           └── rhs: and { lhs: > { lhs: 24, rhs: l_quantity }, rhs: >= { lhs: l_discount, rhs: 0.07 } }
            ├── cost: 7210.72                                                                                       
            └── Scan { cost: 4000, list: [ l_quantity, l_extendedprice, l_discount, l_shipdate ], table: lineitem }
```

Demo:

```
> explain create table t(v1 int not null, v2 int not null);
+---------------------------------------------------------------------------------+
|                                                                                 |
| CreateTable                                                                     |
| ├── columns: [ Column { name: v1, type: INT }, Column { name: v2, type: INT } ] |
| ├── cost: 0.1                                                                   |
| ├── database_id: 0                                                              |
| ├── name: t                                                                     |
| ├── ordered_ids: []                                                             |
| └── schema_id: 0                                                                |
|                                                                                 |
+---------------------------------------------------------------------------------+
in 0.025s
> create table t(v1 int not null, v2 int not null);
+---+
| 1 |
+---+
in 0.003s
> explain insert into t values(1, 1), (4, 2), (3, 3), (10, 12), (2, 5);
+-------------------------------------------------+
|                                                 |
| Insert { cols: [ v1, v2 ], cost: 20, table: t } |
| └── Values { cost: 10, rows: 5 }                |
|     ├── [ 1, 1 ]                                |
|     ├── [ 4, 2 ]                                |
|     ├── [ 3, 3 ]                                |
|     ├── [ 10, 12 ]                              |
|     └── [ 2, 5 ]                                |
|                                                 |
+-------------------------------------------------+
in 0.006s
[...]

PR #778, opened 2023-02-21
feat(pretty): fix singleton lists pretty printing

ice1000:
Was introduced in #771, fix the problem mentioned in https://github.com/risinglightdb/risinglight/pull/777#issuecomment-1435994648


wangrunji0408 (approved):
Thanks

PR #785, opened 2023-05-16
fix: complete the implementaion of `stringconcat`

silver-ymz:
The implementation of `stringconcat` hasn't been completed. It will raise error `execute error: conversion error: no function ||(Utf8, Utf8)`, which needs to add `concat` function in `binary_op`.

Also, could I ask the reason of `O::Item: Sized` limit in `binary_op`? Could I remove it?

### DIFF
--- a/src/array/ops.rs
+++ b/src/array/ops.rs
@@ -206,6 +206,16 @@ impl ArrayImpl {
         }))))
     }
 
+    pub fn concat(&self, other: &Self) -> Result<Self, ConvertError> {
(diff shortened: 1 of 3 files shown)


skyzh:
Yes of course we can remove it. I guess it was introduced to make it easier to implement some functions in the past.

skyzh (approved):
LGTM, thanks a lot!

wangrunji0408:
Stringconcat was added to the planner casually and was never fully supported. Thanks for your efforts to make it work!

PR #790, opened 2023-07-29
chore(deps): update dependencies

wangrunji0408:
Also added a new feature "python" to build pyo3.

### DIFF
--- a/src/binder/delete.rs
+++ b/src/binder/delete.rs
@@ -3,12 +3,13 @@ use super::*;
 impl Binder {
     pub(super) fn bind_delete(
         &mut self,
-        table_name: TableFactor,
+        tables: Vec<ObjectName>,
         selection: Option<Expr>,
     ) -> Result {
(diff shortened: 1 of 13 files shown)


xxchan (approved):

src/binder/delete.rs
```diff
@@ -3,12 +3,13 @@ use super::*;
 impl Binder {
     pub(super) fn bind_delete(
         &mut self,
-        table_name: TableFactor,
+        tables: Vec<ObjectName>,
```
xxchan, line 6: This seems wrong. Guess the vec is dot separated name

wangrunji0408: Nope. It is multiple tables. 😇
https://docs.rs/sqlparser/0.36.1/sqlparser/ast/enum.Statement.html#variant.Delete

xxchan: crazy 🤡

PR #793, opened 2023-08-05
fix: avoid div by zero error when vectorization

### LINKED ISSUE #773: expr: avoid div by zero error when vectorization
```
> create table t2(x int, y int);
+---+
| 1 |
+---+
in 0.045s
> insert into t2 values (1, 2), (2, null);
2023-02-08T05:38:29.126526Z  INFO risinglight::storage::secondary::transaction: RowSet #1 flushed, DV  flushed
+---+
| 2 |
+---+
in 0.050s
> select x / y from t2;
thread 'tokio-runtime-worker' panicked at 'attempt to divide by zero', /rustc/c1a859b25a95999ba276075bbd8e284ea675b41a/library/core/src/ops/arith.rs:488:1
++
++
in 0.018s
```

as `y = null` should yield null, it shouldn't be causing div by zero error.

One possible solution is to find if there's a way to div without exception.

jetjinser:
modify `binary_op` so that op(`F`) only maps to non-empty array items, else `Default::default()`.
add `binary_lop`(binary logic operator), renamed from original `binary_op`

Actually I don't quite understand the vectorization here... I just found that doing it this way prevents the bug mentioned in the issue. Maybe it's wrong to zip like this, or lose the ability to auto-vectorize?

Thanks if anyone can review it.

fix #773.

### DIFF
--- a/src/planner/rules/expr.rs
+++ b/src/planner/rules/expr.rs
@@ -118,7 +118,7 @@ pub fn eval_constant(egraph: &EGraph, enode: &Expr) -> ConstValue {
         Some(x(*e)?.clone())
     } else if let Some((op, a, b)) = enode.binary_op() {
         let (a, b) = (x(a)?, x(b)?);
-        if a.is_null() || b.is_null() || b.is_zero() {
+        if a.is_null() || b.is_null() {
             return Some(DataValue::Null);
         }
         let array_a = ArrayImpl::from(a);
--- a/src/planner/rules/expr.rs
+++ b/src/planner/rules/expr.rs
@@ -118,7 +118,7 @@ pub fn eval_constant(egraph: &EGraph, enode: &Expr) -> ConstValue {
         Some(x(*e)?.clone())
     } else if let Some((op, a, b)) = enode.binary_op() {
         let (a, b) = (x(a)?, x(b)?);
-        if a.is_null() || b.is_null() {
+        if a.is_null() || b.is_null() || b.is_zero() {
             return Some(DataValue::Null);
         }
         let array_a = ArrayImpl::from(a);
--- a/tests/sql/nullable_operator.slt
+++ b/tests/sql/nullable_operator.slt
@@ -1,0 +1,14 @@
+statement ok
+create table t(x int, y int)
+
+statement ok
+insert into t values (1, 2), (2, NULL)
+
+query I
+select x / y from t
+----
+0
+NULL
+
+statement ok
+drop table t
--- a/tests/sql/nullable_operator.slt
+++ b/tests/sql/nullable_operator.slt
@@ -10,5 +10,11 @@ select x / y from t
 0
 NULL
 
+query I
+select x / 0 from t
+----
+NULL
+NULL
+
 statement ok
 drop table t
--- a/src/array/ops.rs
+++ b/src/array/ops.rs
@@ -159,7 +159,7 @@ impl ArrayImpl {
         let (A::Bool(a), A::Bool(b)) = (self, other) else {
             return Err(ConvertError::NoBinaryOp("and".into(), self.type_string(), other.type_string()));
         };
-        let mut c: BoolArray = binary_op(a.as_ref(), b.as_ref(), |a, b| *a && *b);
+        let mut c: BoolArray = binary_lop(a.as_ref(), b.as_ref(), |a, b| *a && *b);
         let a_false = a.to_raw_bitvec().not_then_and(a.get_valid_bitmap());
         let b_false = b.to_raw_bitvec().not_then_and(b.get_valid_bitmap());
         c.get_valid_bitmap_mut().or(&a_false);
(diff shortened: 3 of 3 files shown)


wangrunji0408 (commented):
Hi! Thanks for your interest on this issue.

As you mentioned, this PR breaks auto-vectorization because it requires simple loops with no branches inside. I have run a micro-benchmark to prove this regression.

```sh
cargo bench --bench array -- --quick
```
The main branch:
```
add(i32,i32)/1          time:   [51.212 ns 51.511 ns 51.586 ns]
add(i32,i32)/16         time:   [51.978 ns 51.988 ns 51.990 ns]
add(i32,i32)/256        time:   [68.170 ns 68.226 ns 68.240 ns]
add(i32,i32)/4096       time:   [327.09 ns 334.94 ns 336.90 ns]
```
This PR:
```
add(i32,i32)/1          time:   [67.978 ns 68.130 ns 68.168 ns]
                        change: [+31.775% +32.440% +33.110%] (p = 0.10 > 0.05)
add(i32,i32)/16         time:   [80.365 ns 80.506 ns 80.541 ns]
                        change: [+54.576% +54.765% +54.953%] (p = 0.10 > 0.05)
add(i32,i32)/256        time:   [303.98 ns 304.29 ns 305.56 ns]
                        change: [+345.45% +346.84% +348.23%] (p = 0.10 > 0.05)
add(i32,i32)/4096       time:   [3.9860 µs 3.9885 µs 3.9983 µs]
                        change: [+1083.1% +1102.5% +1122.4%] (p = 0.10 > 0.05)
```

Therefore, I suggest to keep the function `binary_op` unchanged and use the new function only for division. 

In fact, this PR doesn't resolve the problem completely. If we divide a value by zero, it will still panic.

```
> select 1/0;
thread 'main' panicked at 'attempt to divide by zero'
```

In this case, we should either return an error (like postgres) or return a null value (like duckdb). 

```
$ duckdb
v0.8.1 6536a77232
D select 1 / 0;
┌─────────┐
│ (1 / 0) │
│ double  │
├─────────┤
│         │
└─────────┘

$ psql
postgres=# select 1 / 0;
ERROR:  division by zero
```

jetjinser:
Thanks for the reply and explanation, I'd like to move forward with this pr.
Following your suggestion, my idea is to replace all items marked as null in the dividend array with 1's before performing the division. Equivalent to manual mask? But the problem I'm having is that ArrayBuilder's push method fills the empty items with default values, and the default values for numbers are all 0, so I can't construct a valid array through the builder's exposed methods. I don't dare to change the api at will, so I ask you for some suggestions.
Hope it didn't bother you too much, looking forward to your opinion

wangrunji0408:
I'd suggest first building a `try_binary_op` (see [`try_unary_op`](https://github.com/risinglightdb/risinglight/blob/424db3f0adc227a081005908720234329146ec67/src/array/ops.rs#L621) for reference) and calling `checked_div` in it to perform the division. Obviously this is bad in performance, but it's clearly correct. Next, we may add a fast path to vectorize the evaluation in special cases. Like, if all dividends are non-null and non-zero, we can perform the division using `binary_op`.

jetjinser:
thank you for your reply. 

Before seeing your reply, I implemented this function according to the idea I said before. The implementation method is different from what you said, but I think it should be feasible.

What do you think, feel free to comment if it's not appropriate and I'll reconsider the implementation in light of your comments.

`cargo bench --bench array -- --quick`:

```
add(i32,i32)/1          time:   [133.36 ns 137.01 ns 137.93 ns]
                        change: [+11.682% +16.784% +22.181%] (p = 0.10 > 0.05)
                        No change in performance detected.
add(i32,i32)/16         time:   [119.39 ns 120.86 ns 121.22 ns]
                        change: [+0.4522% +1.2453% +2.0388%] (p = 0.11 > 0.05)
                        No change in performance detected.
add(i32,i32)/256        time:   [147.77 ns 148.25 ns 148.36 ns]
                        change: [-4.5686% -2.5708% -0.4962%] (p = 0.13 > 0.05)
                        No change in performance detected.
add(i32,i32)/4096       time:   [891.25 ns 894.66 ns 908.28 ns]
                        change: [+7.5578% +8.9600% +10.372%] (p = 0.10 > 0.05)
                        No change in performance detected.
```

In addition, I also fixed the division by 0 error when selecting constants, and now the behavior is consistent with duckdb (the result is NULL)

src/planner/rules/expr.rs
```diff
@@ -118,7 +118,7 @@ pub fn eval_constant(egraph: &EGraph, enode: &Expr) -> ConstValue {
         Some(x(*e)?.clone())
     } else if let Some((op, a, b)) = enode.binary_op() {
         let (a, b) = (x(a)?, x(b)?);
-        if a.is_null() || b.is_null() {
+        if a.is_null() || b.is_null() || b.is_zero() {
```
wangrunji0408, line 121: should be `b.is_zero() && op == Div`. otherwise any `a + 0` or `a - 0` would result in NULL.
However, I believe this condition is unnecessary if array division is implemented correctly.

tests/sql/nullable_operator.slt
wangrunji0408, line 1: better to add tests for `x / 0`:

```slt
query I
select x / 0 from t;
----
NULL
NULL
```

I guess this test will fail because you forgot to handle zero in array operation :)

wangrunji0408 (commented):
I think your solution is feasible. Thanks!

tests/sql/nullable_operator.slt
jetjinser: you are right i forgot it 😢

jetjinser:
It should work fine now i think 😣

wangrunji0408 (approved):
LGTM, Thanks!

jetjinser:
Thanks for the quick review. Coincidentally, I just made a little style change 😂

I'm happy to be able to contribute to this project 🥰

wangrunji0408:
Welcome to join the community!

PR #795, opened 2023-08-11
fix(optimizer): support pushdown pk to storage scan

skyzh:
Add a new rule that maps null to true so that storage filter on pk can be pushed down.

### DIFF
--- a/src/planner/rules/range.rs
+++ b/src/planner/rules/range.rs
@@ -106,11 +106,11 @@ fn is_primary_key_range(expr: &str) -> impl Fn(&mut EGraph, Id, &Subst) -> bool 
         let Some((column, _)) = &egraph[subst[var]].data.range else {
             return false;
         };
-        egraph
-            .analysis
-            .catalog
-            .get_column(column)
-            .unwrap()
-            .is_primary()
+        if let Some(col) = egraph.analysis.catalog.get_column(column) {
+            col.is_primary()
+        } else {
+            // handle the case that catalog is not initialized, like in test cases
+            false
+        }
     }
 }
--- a/tests/planner_test/storage-pushdown.planner.sql
+++ b/tests/planner_test/storage-pushdown.planner.sql
@@ -1,0 +1,57 @@
+-- use merge join for primary key joins
+explain select * from t1 join t2 on a = c;
+
+/*
+Join { type: inner, on: = { lhs: c, rhs: a }, cost: 0, rows: 0 }
+├── Scan { table: t1, list: [ a, b ], filter: true, cost: 0, rows: 0 }
+└── Scan { table: t2, list: [ c, d ], filter: true, cost: 0, rows: 0 }
+*/
+
+-- use storage order by instead of sorting by primary key
+explain select * from t1 order by a;
+
+/*
+Scan { table: t1, list: [ a, b ], filter: true, cost: 0, rows: 0 }
+*/
+
+-- use storage filter for primary key
+explain select * from t1 where a = 1;
+
+/*
+Scan { table: t1, list: [ a, b ], filter: = { lhs: a, rhs: 1 }, cost: 10, rows: 5 }
+*/
+
+-- use storage filter for a combination of primary key and other keys
+explain select * from t1 where a > 1 and a < 3 and b > 1;
+
+/*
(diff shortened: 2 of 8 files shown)


skyzh:
it seems that `filter-scan-1` is still not correctly applied, that the last query in the planner test is not pushed down?

skyzh:
fixed b/c we didn't insert data into the table

skyzh:
... and obviously there are some sub-optimal plans 🤪 like two `a > x` and always false condition not eliminated.

wangrunji0408:
> fixed b/c we didn't insert data into the table

Yes, in CBO, if a table is empty, then any plan is possible because all of them have zero cost...

src/planner/rules/range.rs
```diff
@@ -87,6 +87,10 @@ pub fn analyze_range(egraph: &EGraph, enode: &Expr) -> RangeCondition {
 #[rustfmt::skip]
 pub fn filter_scan_rule() -> Vec<Rewrite> { vec![
     // pushdown range condition to scan
+    rw!("filter-scan-null-eq-true";
+        "(filter ?cond (scan ?table ?columns null))" =>
+        "(filter ?cond (scan ?table ?columns true))"
+    ),
```
wangrunji0408, line 93: Should we replace all empty conditions from `null` to `true` again? Because filtering on null is actually equivalent to filtering on false...

skyzh: hmm... really? what's the difference?

I thought null == passing None to the storage iterator filter, and inside the storage implementation null == true.

wangrunji0408: Think about we are going to estimate the number of rows for the scan node:
```
(scan ?table ?columns ?filter)
```

The formula should be:
```
row_count(?table) * selectivity(?filter)
```

Here comes the difference:
```
selectivity(null) == 0
selectivity(true) == 1
```

skyzh: oh so I guess we are messing up with the default condition for storage filter. It should be `true` by default, instead of `null` by default?

tests/planner_test/storage-pushdown.planner.sql
```diff
@@ -0,0 +1,57 @@
...
+
+-- use storage filter for a combination of primary key (always false) and other keys
+explain select * from t1 where a > 1 and a < 0 and b > 1;
+
+/*
+Filter { cond: > { lhs: b, rhs: 1 }, cost: 16.05, rows: 2.5 }
+└── Scan
+    ├── table: t1
+    ├── list: [ a, b ]
+    ├── filter: and { lhs: > { lhs: 0, rhs: a }, rhs: > { lhs: a, rhs: 1 } }
```
wangrunji0408, line 45: Seems that we should add some rules to rewrite this to `false` 🤔

Maybe:
```rust
rw!(""; "(and (> ?x ?a) (< ?x ?b))" => "false" if is_greater_than_or_equal("?a", "?b"))
```

tests/planner_test/storage-pushdown.planner.sql
```diff
@@ -0,0 +1,57 @@
...
+    ├── cost: 10
+    └── rows: 5
+*/
+
+-- use storage filter for a combination of primary key (could be eliminated) and other keys
+explain select * from t1 where a > 1 and a > 3 and b > 1;
+
+/*
+Filter { cond: and { lhs: > { lhs: b, rhs: 1 }, rhs: > { lhs: a, rhs: 3 } }, cost: 15.1, rows: 1.25 }
+└── Scan { table: t1, list: [ a, b ], filter: > { lhs: a, rhs: 1 }, cost: 10, rows: 5 }
```
wangrunji0408, line 55: ditto. `a > 1 and a > 3` should be rewrited to `a > 3`.

Maybe:
```rust
rw!(""; "(and (> ?x ?a) (> ?x ?b))" => "(> ?x ?a)" if is_greater_than_or_equal("?a", "?b"))
```

wangrunji0408 (approved):
LGTM!

tests/planner_test/storage-pushdown.planner.sql
skyzh: will make it in a separate pull request

PR #801, opened 2023-10-16
chore: replace read_to_end()

xiaguan:
Use a more simple interface from tokio::fs::read.


skyzh (approved):

skyzh:
would you please run `cargo fmt`? thanks!

PR #804, opened 2023-11-13
(fix/doc): fix error link

caicancai:
The project code seems to have undergone a refactoring and some links are faulty. Optimizer links are also wrong, but I don't know which one to fill in, I can't find the corresponding module, it seems to go with the planner module

### DIFF
--- a/docs/03-architecture-overview.md
+++ b/docs/03-architecture-overview.md
@@ -173,7 +173,7 @@ In database systems, there are a lot of logical operators. Here we list some:
 * Join: join results from two executors. For example, `select * from a inner join b`.
 * ...
 
-The logical planner will do some simple mappings from bound statements to logical plans, so as to have a basic sketch of how this query will be processed. See [`src/planner`](../src/logical_planner) for more information.
+The logical planner will do some simple mappings from bound statements to logical plans, so as to have a basic sketch of how this query will be processed. See [`src/logical_planner`](../src/planner) for more information.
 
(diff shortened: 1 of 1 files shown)


docs/03-architecture-overview.md
```diff
@@ -173,7 +173,7 @@ In database systems, there are a lot of logical operators. Here we list some:
 * Join: join results from two executors. For example, `select * from a inner join b`.
 * ...
 
-The logical planner will do some simple mappings from bound statements to logical plans, so as to have a basic sketch of how this query will be processed. See [`src/logical_planner`](../src/logical_planner) for more information.
+The logical planner will do some simple mappings from bound statements to logical plans, so as to have a basic sketch of how this query will be processed. See [`src/logical_planner`](../src/planner) for more information.
```
wangrunji0408, line 176:
```suggestion
The logical planner will do some simple mappings from bound statements to logical plans, so as to have a basic sketch of how this query will be processed. See [`src/planner`](../src/planner) for more information.
```

wangrunji0408 (approved):
Thanks for the fix!
In fact, I think the entire article about planner needs a rewrite after the refactoring. But we can do it step by step.

caicancai:
> Thanks for the fix! In fact, I think the entire article about planner needs a rewrite after the refactoring. But we can do it step by step.

Yes, it seems that the whole project has been refactored and the documentation is no longer suitable

PR #806, opened 2023-11-15
chore/catalog: add root test and fix doc

caicancai:
2023-11-15 14-59-24屏幕截图
Document description error

### DIFF
--- a/src/catalog/root.rs
+++ b/src/catalog/root.rs
@@ -158,8 +158,8 @@ mod tests {
         assert_eq!(schema_catalog1.name(), DEFAULT_SCHEMA_NAME);
 
         let schema_catalog2 = catalog.get_schema_by_name(DEFAULT_SCHEMA_NAME).unwrap();
-        assert_eq!(schema_catalog2.id(), 0);
-        assert_eq!(schema_catalog2.name(), DEFAULT_SCHEMA_NAME);
+        assert_eq!(schema_catalog1.id(), schema_catalog2.id());
+        assert_eq!(schema_catalog1.name(), schema_catalog2.name());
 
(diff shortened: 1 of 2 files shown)


caicancai:
I also found some bugs with functions such as sum and am trying to fix them

src/catalog/root.rs
```diff
@@ -143,3 +143,28 @@ fn split_name(name: &str) -> Option<(&str, &str)> {
...
+
+        let schema_catalog2 = catalog.get_schema_by_name(DEFAULT_SCHEMA_NAME).unwrap();
+        assert_eq!(schema_catalog1.id(), schema_catalog2.id());
+        assert_eq!(schema_catalog1.name(), schema_catalog2.name());
+
+        let col = ColumnCatalog::new(0, DataTypeKind::Int32.not_null().to_column("a".into()));
+        let table_id = catalog
+            .add_table(0, "t".into(), vec![col], false, vec![])
+            .unwrap();
+        assert_eq!(table_id, 0);
```
wangrunji0408, line 168: This seems not related to the root catalog. But it's okay to have it here.

wangrunji0408 (approved):
Thanks!
Previously we have the root-database-schema-table hierarchy. It seems too complex for an embedded database. So the "database" level was removed afterwards.

src/catalog/root.rs
caicancai: The main purpose here is to validate the addTable API

wangrunji0408: Oh I see.

caicancai:
> 谢谢！以前我们有 root-database-schema-table 层次结构。对于嵌入式数据库来说，这似乎太复杂了。因此，后来删除了“数据库”级别。

I also have such doubts here, thank you for the answer

PR #809, opened 2023-11-15
chore/binder: add test on binder

caicancai:
2023-11-15 18-38-36屏幕截图

### DIFF
--- a/src/binder/create_table.rs
+++ b/src/binder/create_table.rs
@@ -195,9 +195,8 @@ impl From<&ColumnDef> for ColumnCatalog {
 mod tests {
     use std::sync::Arc;
 
-    use super::*;
     use crate::binder::Binder;
-    use crate::catalog::{ColumnCatalog, RootCatalog};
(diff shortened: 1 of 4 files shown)


caicancai:
I'll see if I can fill in the other tests tomorrow

caicancai:
2023-11-16 09-44-04屏幕截图

caicancai:
@wangrunji0408 hi, excuse me. if you have time, could you review my two pr? I can close this pr if there is a problem

skyzh:
will take a look this weekend :)

skyzh (commented):
I don't think these new tests are actually testing + comparing with some expected result? In this case, it looks better to make this test part of sqlplannertest, so that we can find regressions across commits.

caicancai:
> I don't think these new tests are actually testing + comparing with some expected result? In this case, it looks better to make this test part of sqlplannertest, so that we can find regressions across commits.

thank your review and advice

PR #811, opened 2023-11-21
binder/chore: remove drop test

caicancai:
https://github.com/risinglightdb/risinglight/pull/809

### DIFF
--- a/src/binder/drop.rs
+++ b/src/binder/drop.rs
@@ -79,22 +79,3 @@ impl Binder {
         }
     }
 }
-
-#[cfg(test)]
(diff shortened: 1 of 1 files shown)


caicancai:
At present, drop test is not necessary, one is that this code is not used
```rust
let catalog = Arc::new(RootCatalog::new());
        catalog
            .add_table(0, "mytable".into(), vec![], false, vec![])
            .unwrap();
```

caicancai:
cc @wangrunji0408

wangrunji0408 (approved):

PR #813, opened 2023-11-29
chore/rule: add expr test for rule

caicancai:


caicancai:
Attempt to divide by zero can be used as a separate test. I refer to calcite to do the same.

caicancai:
I will complete the todo content in the comments and try to fix this bug.
tidb and mysql also seem to have this bug

caicancai:
I have a question that I would like to ask in this PR. I would like to ask if this situation is considered a bug.
2023-11-30 00-15-56屏幕截图

caicancai:
I feel that the display results of risinglight's select can be improved, and I am trying this part of the work

wangrunji0408 (approved):
Thanks!

caicancai:
> Here is the output of DuckDB:
> 
> ```
> D select cast('8145308033243873281' as double) ;
> ┌───────────────────────────────────────┐
> │ CAST('8145308033243873281' AS DOUBLE) │
> │                double                 │
> ├───────────────────────────────────────┤
> │                 8.145308033243873e+18 │
> └───────────────────────────────────────┘
> ```
> 
> So yes, I also think the current output of RisingLight is not correct.

Thank you for your reply and patient answer. As far as I know, tidb and mysql also have similar bugs. I will try to solve this problem in risinglight. Finally, I would like to thank you for the risinglight project. As a beginner, I have learned a lot. Thank you very much for your contribution.

caicancai:
The work of risinglight community is very meaningful

PR #815, opened 2023-12-02
feat(type): support type-string for timestamp type

caicancai:
I feel like it’s very simple whether timestamp will support it. There may be some problems that I didn’t pay attention to.

### DIFF
--- a/src/binder/expr.rs
+++ b/src/binder/expr.rs
@@ -165,6 +165,14 @@ impl Binder {
                 })?;
                 Ok(self.egraph.add(Node::Constant(DataValue::Date(date))))
             }
+            DataType::Timestamp(_, _) => {
+                let timestamp = value.parse().map_err(|_| {
+                    BindError::CastError(DataValue::String(value), DataTypeKind::Timestamp)
+                })?;
+                Ok(self
+                    .egraph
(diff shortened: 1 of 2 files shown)


caicancai:
I'm thinking about whether I need to support time type

wangrunji0408 (approved):
LGTM!

Actually the timestamp type is already supported.

```
> select '2022-01-01 00:00:00'::timestamp;
+---------------------+
| 2022-01-01 00:00:00 |
+---------------------+
in 0.034s
```

What you have done in this PR is the **type-string support** for timestamp type.

```
timestamp '2022-01-01 00:00:00'
```

Thank you for your contribution! Even though it is a small step.

wangrunji0408:
> I'm thinking about whether I need to support time type

Feel free to have a try. It would be more challenging than this one. 😀

caicancai:
> > I'm thinking about whether I need to support time type
> 
> Feel free to have a try. It would be more challenging than this one. 😀

Thanks, I'm going to try it, I want to ask about the writing of the optimizer I may not understand very well, is this a rust-specific way of writing.
```rust
rw!("inner-join-on-right-empty"; "(join inner ?on ?l (empty ?r))"   => "(empty ?l ?r)"),
```

Issue #820, opened 2023-12-03
`select *` panic in subquery


skyzh:
```
> create table t1(v1 int, v2 int);
> explain select * from (select v1 + v1 as v3 from t1) as left;
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/planner/rules/plan.rs:266:51
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

caicancai:
It's possible to return a default value when capturing None, I'm not sure that's a good idea

caicancai:
> This bug has been fixed in #796. However, it has not been merged for a long time. 🥲 I can't recall too much details for now. But in my mind, this involves a tricky way how we handle the output of subquery. I might write an article to explain it in detail. (

#796 looks very cool, looking forward to the author’s article

wangrunji0408:
fixed by #796

PR #821, opened 2023-12-05
fix: select * panic in subquery

caicancai:
close: #820


caicancai:
I'm not sure if this is a good idea

Issue #385, opened 2022-01-30
catalog: refactor `ColumnCatalog` and `ColumnDesc`


st1page:
now we have the the definitions:
```rust 
pub struct ColumnDesc {
    datatype: DataType,
    is_primary: bool,
}
pub struct ColumnCatalog {
    id: ColumnId,
    name: String,
    desc: ColumnDesc,
}
```
we need do the following refactor:
1. remove the `is_primiary` from the `ColumnDesc`, and add the `primiary_column: ColumnId` in `SecondaryTable` and `TableCatalog` 
2. move the `ColumnCatalog::name` into `ColumnDesc::name`

Issue #384, opened 2022-01-29
storage: support multiple primary keys


skyzh:

st1page:
we might need to do the refactor first #385. because after we have composite primary keys, the `is_primary` field of a column is weird and confusing.

PR #823, opened 2023-12-09
feat(storage): Support composite primary keys

danipozo:
Aims to solve #384, #385.

Still need to handle composite primary keys in `BTreeMapMemTable`, submitting for early feedback as there may be some other things I've got wrong about the code base.

### DIFF
--- a/src/catalog/column.rs
+++ b/src/catalog/column.rs
@@ -126,17 +126,14 @@ impl ColumnCatalog {
 }
 
 /// Find the id of the sort key among column catalogs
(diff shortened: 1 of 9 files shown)


caicancai:
@danipozo Maybe you need to implement cargo fmt

skyzh (approved):

skyzh:
would you please rebase to main + fmt? thanks a lot!

danipozo:
> would you please rebase to main + fmt? thanks a lot!

Done!

PR #827, opened 2024-01-13
feat(catalog): add function catalog

xzhseh:
As titled, as a preparation for sql (and maybe more...) udf.

Currently `FunctionCatalog` stores fairly simple things: 
- function name
- argument types
- return type of the function
- the language used by function definition
- function body (i.e., the actual definition)

cc @wangrunji0408.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /.vscode
 .gdb_history
 risinglight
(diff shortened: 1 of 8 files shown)


wangrunji0408 (approved):
LGTM!

PR #828, opened 2024-01-13
feat(function): add `CreateFunctionExecutor` & `bind_create_function`

xzhseh:
As titled, cc @wangrunji0408.

P.S. Some of the logic in `bind_create_function` is the same as `create_sql_function`. 😄


wangrunji0408 (approved):
LGTM. Thanks!

PR #829, opened 2024-01-28
feat(sql-udf): add support for named sql udf

xzhseh:
We now support named sql udf in risinglight!

e.g., the following definitions are totally valid and could be used as normal built-in functions now.
```sql
create function add_named(a INT, b INT) returns int language sql as 'select a + b';
```

```sql
create function sub_named(a INT, b INT) returns int language sql as 'select a - b';
```

See the test cases for detailed use cases.

P.S. anonymous sql udf may be supported in subsequent PRs, with other necessary preparations.

cc @wangrunji0408.


src/binder/mod.rs
```diff
@@ -97,6 +97,117 @@ pub struct Binder {
     contexts: Vec<Context>,
     /// The number of occurrences of each table in the query.
     table_occurrences: HashMap<TableRefId, u32>,
+    /// The context used in sql udf binding
+    udf_context: UdfContext,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct UdfContext {
+    /// The mapping from `sql udf parameters` to a bound `ExprImpl` generated from `ast
```
xzhseh, line 106:
```suggestion
    /// The mapping from `sql udf parameters` to a bound `Node` generated from `ast
```

wangrunji0408 (commented):
LGTM. But I prefer to complete a minimal workable SQL UDF in a single PR. This way we can revisit the whole design and see where can be optimized or simplified. 🤔

xzhseh:
> LGTM. But I prefer to complete a minimal workable SQL UDF in a single PR. This way we can revisit the whole design and see where can be optimized or simplified. 🤔

For sure, I'll add the anonymous sql udf support in this PR later.

PR #832, opened 2024-02-24
refactor: remove `Internal` node and add more system catalogs

wangrunji0408:
- remove `Internal` node from plan, merged into `Scan` node
- replace `InternalTableExecutor` by `SystemTableScanExecutor`
- add new system tables: `pg_tables`, `pg_attribute` and `pg_stat`
- expand commands like `\dt` to scan from system tables
- ~~remove hardcoded contributor list, fetch from Github instead~~ update contributor list


skyzh (approved):

tests/sql/internal.slt
```diff
@@ -1,108 +1,7 @@
 query I
-select * from pg_catalog.contributors;
+select count(github_id) > 50 from pg_catalog.contributors;
```
skyzh, line 2: note that GitHub as request limits for such things, maybe we need to exclude it from CI?

wangrunji0408: Hmmm, looks like it makes more trouble than hardcoding. Let me revert that.

PR #833, opened 2024-02-26
refactor(test): move sqllogictest and sqlplannertest to single file tests

wangrunji0408:
... so that these tests can be included when running `cargo test`. Previously they were easily forgotten to run locally.

Additionally, this PR adds a bench for TPC-H: `cargo bench --bench tpch`


xxchan (approved):
LGTM

PR #835, opened 2024-03-02
feat(binder): support sql udf with unnamed parameters

xzhseh:
A minimal support for sql udf with unnamed parameters.


wangrunji0408 (approved):
LGTM!

PR #836, opened 2024-03-07
refactor(type): move nullable out of `DataType`

wangrunji0408:
This PR moves `nullable` out of `DataType` and puts it into `ColumnDesc`. The original `DataTypeKind` is renamed to `DataType`.

This is to make it align with other projects like arrow-rs and RisingWave. Actually I'm planning to migrate the entire array module to arrow-rs, so that we can be more standard and have better interoperability with the community.


skyzh (approved):
Thanks!

PR #837, opened 2024-03-12
feat: support view in memory

wangrunji0408:
This PR is a part of #796, adds support for creating, querying and dropping views in memory.

The key implementations are: 
1. When creating a view, bind the query and store the logical plan with the view in catalog.
2. When querying from a view, build executors for all views and then build other plan nodes on top of them. Given that a view can be consumed by multiple downstream nodes, we introduce `StreamSubscriber` to allow multiple consumers of a stream.

Limitations:
1. We don't persist views in disk storage.
2. We don't support inferring schema from the query. Columns must be defined explicitly when creating a view.
3. We don't maintain dependency relationship between tables and views.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -119,12 +119,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
 
 [[package]]
+name = "async-broadcast"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "258b52a1aa741b9f09783b2d86cf0aeeb617bbf847f6933340a39644227acbdb"
+dependencies = [
+ "event-listener 5.2.0",
+ "event-listener-strategy",
+ "futures-core",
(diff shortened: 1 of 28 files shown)


skyzh:
> When querying from a view, build executors for all views and then build other plan nodes on top of them. Given that a view can be consumed by multiple downstream nodes, we introduce StreamSubscriber to allow multiple consumers of a stream.

Well, then we have DAG in the system 🤪 I thought an easier way is just to create multiple copies of the plan and then execute it multiple times.

wangrunji0408:
> > When querying from a view, build executors for all views and then build other plan nodes on top of them. Given that a view can be consumed by multiple downstream nodes, we introduce StreamSubscriber to allow multiple consumers of a stream.
> 
> Well, then we have DAG in the system 🤪 I thought an easier way is just to create multiple copies of the plan and then execute it multiple times.

I think it is natural to have DAG in data processing pipelines. It'd be better to reuse the result from common upstreams. Another interesting fact is that a query plan in egg's e-graph is also a DAG, even if you don't construct DAG intentionally. Because egg can automatically identify and merge equal nodes. This will make it easier to eliminate common subexpressions and even reuse CTEs in the future.

PR #838, opened 2024-03-15
feat: support WITH clause

wangrunji0408:
This PR adds support for specifying CTEs using the `WITH` clause. We treat CTEs just as named subqueries. The implementation simply adds CTE queries to the binder context and inlines them when referenced.

However, due to an optimizer bug (which has been fixed in #796), the current CTE is pretty useless as most practice use cases will lead to panics.


TennyZhuang (approved):
LGTM!

PR #839, opened 2024-03-15
feat: use ahash for better hash performance

xiaguan:
For tpch 10GB 's Q6 , 1:11 -> 1:05


xiaguan:
`std::collections::HashMap` provides strong guarantees regarding DOS (Denial of Service) resistance, while `ahash` prioritizes performance over such guarantees.

PR #841, opened 2024-03-16
feat: parallelism the hash aggregation

xiaguan:
For tpch 10gb q1 , which has a lot of aggregation.
1:10 -> 40s
It's not a pull request indeed.
I think we need a paralleism executor.


wangrunji0408:
Good experiment! But I think the final solution should be building the same executors for each core. The executor itself should focus on processing data in a sequential manner.

Issue #834, opened 2024-02-28
optimizer: filter scan rule


crwen:
https://github.com/risinglightdb/risinglight/blob/0e4a8469e74d6e8a468b58e82df9ac0d0311b5e3/src/planner/rules/range.rs#L88-L100

Here's my understanding: When there are predicates on primary key, we should push them down to the storage layer. Is that right?

But it seems not to work when I execute statements like `select pk from t where pk > 1`. And the [`filter`](https://github.com/risinglightdb/risinglight/blob/0e4a8469e74d6e8a468b58e82df9ac0d0311b5e3/src/executor/table_scan.rs#L44) passed to [`scan`](https://github.com/risinglightdb/risinglight/blob/0e4a8469e74d6e8a468b58e82df9ac0d0311b5e3/src/storage/memory/transaction.rs#L121) function is always `None`. It looks like these:
```
> explain select a from t2 where a > 17;
Filter { cond: > { lhs: a, rhs: 17 }, cost: 68.4, rows: 20 }
└── Scan { table: t2, list: [ a ], filter: null, cost: 40, rows: 40 }
```

Maybe the `true` in `"(scan ?table ?columns true)"` should be replaced to `null`? Because it seems to work if I change it.

```
> explain select a from t2 where a > 17;
Scan { table: t2, list: [ a ], filter: > { lhs: a, rhs: 17 }, cost: 40, rows: 40 }
```

skyzh:
maybe related: https://github.com/risinglightdb/risinglight/pull/795? But it's a known issue and if you want to fix that, please submit a pull request, thanks!

crwen:
It seems like that the filter is true by default, instead of null. So the solution may be like this in #795?


https://github.com/risinglightdb/risinglight/blob/cd03496eb2c0cb1fab7e43fa5738d42c091013fd/src/binder/table.rs#L60-L70

Issue #794, opened 2023-08-11
optimizer: push any filter down to storage


skyzh:
The current storage iterator supports pushing down any filter. We may add an optimizer rule for that (I myself is working on it...)

wangrunji0408:
What's the difference between filtering arbitrary expression on the executor and filtering on the storage? Is that really helpful to push it down? I can only think of range filter on primary key being more efficient. That's why I didn't add rules to push any filter down. 🤪

skyzh:
Filters with high selectivity can also be super efficient if it's pushed down to the storage side. Consider two examples:

1. The current storage engine supports conditional read. That is to say, if we have something like `select * from t where a > 10 and b < 10` even if they are not pks, we will first read a block of column a and column b, and if all data in that block does not match the condition, we will not read the remaining columns and will move the cursor.
2. Storage (may) support filtering by metadata, whereas we store min/max for each of the 1MB storage block. Therefore, if we have a high selectivity filter like `b=10` on any column, we should be able to filter a lot of things.

wangrunji0408:
I see. Thanks for the explanation. Then we should revert some changes from #786, where I removed the code for filtering in storage. 🤪

PR #842, opened 2024-03-17
fix(optimizer): fix filter pushdown and range fold

kysshsy:
related  to #794 #834 , add some new rules and change scan's default filter to true(this make it match with the "filter-scan" rule)

### DIFF
--- a/src/planner/rules/expr.rs
+++ b/src/planner/rules/expr.rs
@@ -73,6 +73,18 @@ pub fn rules() -> Vec<Rewrite> { vec![
     rw!("and-null";  "(and null ?a)"    => "null"),
     rw!("and-comm";  "(and ?a ?b)"      => "(and ?b ?a)"),
     rw!("and-assoc"; "(and ?a (and ?b ?c))" => "(and (and ?a ?b) ?c)"),
+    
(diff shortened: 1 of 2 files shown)


src/planner/rules/expr.rs
```diff
@@ -172,6 +184,47 @@ fn is_not_zero(var: &str) -> impl Fn(&mut EGraph, Id, &Subst) -> bool {
...
+    f: impl Fn(&DataValue, &DataValue) -> bool,
+) -> impl Fn(&mut EGraph, Id, &Subst) -> bool {
+    let v1 = var(v1);
+    let v2 = var(v2);
+
+    move |egraph, _, subst| match (
+        &egraph[subst[v1]].data.constant,
+        &egraph[subst[v2]].data.constant,
+    ) {
+        (Some(d1), Some(d2)) => f(d1, d2),
```
wangrunji0408, line 223: Seems like the common condition `std::mem::discriminant(d1) == std::mem::discriminant(d2)` can be moved to here.

wangrunji0408 (approved):
LGTM. Thank you!

Should we get #795 merged first? @skyzh

skyzh:
yep i'll handle these two pull requests on the weekends :)

skyzh:
need rebasing + fix conflicts, I can help with that next weekend :)

kysshsy:
rebase the commit, help me check the sql test result, thank you 😁

PR #847, opened 2024-04-16
fix(ci): fix benchmark workflow

wangrunji0408:
fix some bugs in #845 😅


skyzh (approved):

wangrunji0408:
> > Error: The operation was canceled.
> 
> either timeout or someone cancelled it?

I suppose it was killed because of OOM when running Q21. (it consumes over 20GB memory on my Mac)

let me skip q21 for now.

PR #851, opened 2024-06-01
chore: add rule and test

caicancai:
I noticed that risinglight uses egg for CBO optimization, but it looks violent at the moment. I will try to optimize it.

### DIFF
--- a/src/planner/rules/plan.rs
+++ b/src/planner/rules/plan.rs
@@ -618,7 +618,7 @@ mod tests {
             )
         )" => "
         proj (list $1.2)
-        （topn 10 0 (list) 
+        （topn 10 0 (list $1.2) 
             (scan $1 (list $1.1 $1.2) null)
         )
         "
--- a/src/planner/rules/plan.rs
+++ b/src/planner/rules/plan.rs
@@ -613,7 +613,7 @@ mod tests {
         "
         proj (list $1.2)
         (limit 10 0
-            (order (list) 
+            (order (list $1.2) 
                 (scan $1 (list $1.1 $1.2) null)
             )
         )" => "
--- a/src/planner/rules/plan.rs
+++ b/src/planner/rules/plan.rs
@@ -21,6 +21,7 @@ fn cancel_rules() -> Vec<Rewrite> { vec![
     rw!("limit-null";       "(limit null 0 ?child)"     => "?child"),
     rw!("order-null";       "(order (list) ?child)"     => "?child"),
     rw!("filter-true";      "(filter true ?child)"      => "?child"),
+    rw!("filter-false";     "(filter false ?child)"     => "(empty ?child)"),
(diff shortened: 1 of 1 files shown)


caicancai:
cc @wangrunji0408 If you have time, could you help me review this PR? Thank you.

src/planner/rules/plan.rs
```diff
@@ -582,6 +583,47 @@ mod tests {
...
+        merge_filter,
+        rules(),
+        // SELECT name
+        // FROM student
+        // LIMIT 10
+        // Order by name
+        "
+        proj (list $1.2)
+        (limit 10 0
+            (order (list)
```
wangrunji0408, line 616: This is an empty ordering. Should be `(order (list $1.2) ..)`?

caicancai: Thanks, fix
But the fact that rust test can pass means there is still room for improvement. I will find time to improve it.

caicancai:
@wangrunji0408 PATL，thank you

src/planner/rules/plan.rs
```diff
@@ -582,6 +583,47 @@ mod tests {
...
+        // Order by name
+        "
+        proj (list $1.2)
+        (limit 10 0
+            (order (list $1.2) 
+                (scan $1 (list $1.1 $1.2) null)
+            )
+        )" => "
+        proj (list $1.2)
+        （topn 10 0 (list)
```
wangrunji0408, line 621: But why this test can pass?

```suggestion
        （topn 10 0 (list $1.2) 
```

caicancai: I think this is a bug and I will fix it

caicancai:
@wangrunji0408 I would like to ask a question, can egg+sql optimization be extracted into a separate framework and then integrated with calcite? Do you think this is feasible?

src/planner/rules/plan.rs
caicancai: cli failed.
Why?

wangrunji0408: Permission denied. It's a random happened failure. Let's retry it.

wangrunji0408:
> can egg+sql optimization be extracted into a separate framework and then integrated with calcite? Do you think this is feasible?

I'm not sure as the Calcite is written in Java while egg is designed for Rust. Besides, what's the purpose of doing this? As far as I know, the Calcite itself is a framework for optimization. I have always thought of egg as a Rust alternative of Calcite.

caicancai:
> I'm not sure as the Calcite is written in Java while egg is designed for Rust. Besides, what's the purpose of doing this? As far as I know, the Calcite itself is a framework for optimization. I have always thought of egg as a Rust alternative of Calcite.

I think it is a bit complicated to expand an optimization rule in Calcite at present, and it is not as convenient as egg. My original intention is to make Calcite more convenient to use.

wangrunji0408 (approved):
Thanks!

PR #852, opened 2024-06-15
fix: limit 0 return empty

caicancai:


skyzh (commented):
LGTM, but it seems that a lot of regress tests are failing because we don't populate data for these queries. Need to figure out a way to fix them before merging. Thanks anyways!

caicancai:
> LGTM, but it seems that a lot of regress tests are failing because we don't populate data for these queries. Need to figure out a way to fix them before merging. Thanks anyways!

I'll try to fix it over the weekend.
Do we just need to fill in the data of the test sample? I am not very clear about this part. Can you give me some advice?

src/planner/rules/plan.rs
```diff
@@ -18,7 +18,7 @@ pub fn always_better_rules() -> Vec<Rewrite> {
 
 #[rustfmt::skip]
 fn cancel_rules() -> Vec<Rewrite> { vec![
-    rw!("limit-null";       "(limit null 0 ?child)"     => "?child"),
+    rw!("limit-null";       "(limit null 0 ?child)"     => "(empty ?child)"),
```
wangrunji0408, line 21: The definition of the limit node is `(limit limit_num offset_num child)`.
So `(limit null 0 ?child)` means `LIMIT ALL OFFSET 0`, which is a dummy operation.
Actually this is not a bug. 😄

wangrunji0408 (approved):

src/planner/rules/plan.rs
caicancai: Thanks for the reminder, but limit 0 actually returns a logical plan similar to
`LogicalValues(tuples=[[]]).`
 I don't know if risinglight has implemented it now. If not, I'd be happy to contribute.

Issue #864, opened 2025-01-08
vector store + integration


skyzh:
- [ ] vector type
- [ ] vector functions
- [ ] vector storage format (use JSON for now?)
- [ ] vector indexes
- [ ] integration with langchain (WIP)

skyzh:
misc:

- [ ] create table if not exists
- [ ] more efficient insert into statements via python interface (passing string is slow)

PR #865, opened 2025-01-08
feat(binder): support create index statements

skyzh:
part of https://github.com/risinglightdb/risinglight/issues/864

Add `create index` support. No indexes are created for now. It only modifies the catalog. All storage engines now use a shared submodule called `index` to manage in-memory indexes.

### DIFF
--- a/src/catalog/root.rs
+++ b/src/catalog/root.rs
@@ -234,10 +234,7 @@ const CREATE_SYSTEM_TABLE_SQL: &str = "
         schema_id int not null,
         schema_name string not null,
         table_id int not null,
-        table_name string not null,
-        index_id int not null,
-        index_name string not null,
-        on_columns string not null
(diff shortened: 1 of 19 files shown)


src/catalog/root.rs
```diff
@@ -206,7 +234,19 @@ const CREATE_SYSTEM_TABLE_SQL: &str = "
         schema_id int not null,
         schema_name string not null,
         table_id int not null,
-        table_name string not null
+        table_name string not null,
+        index_id int not null,
+        index_name string not null,
+        on_columns string not null
```
wangrunji0408, line 240: misoperation?

wangrunji0408 (approved):
LGTM!

src/catalog/root.rs
skyzh: fixed

PR #866, opened 2025-01-08
feat(array,types): support vector type

skyzh:


src/array/bytes_array.rs
```diff
@@ -9,40 +9,59 @@ use bitvec::vec::BitVec;
...
 
 use super::{Array, ArrayBuilder, ArrayEstimateExt, ArrayFromDataExt, ArrayValidExt};
-use crate::types::BlobRef;
+use crate::types::{BlobRef, VectorRef, F64};
 
-/// A collection of variable-length values.
+// A collection of variable-length values.
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
-pub struct BytesArray<T: ValueRef + ?Sized> {
+pub struct BytesArray<T: ValueRef<U> + ?Sized, U: PrimitiveValueType = u8> {
```
xxchan, line 16: Should we rename it? It doesn't look like `Bytes` any more

xxchan: Besides, should Vector have variable-length or fixed length?

wangrunji0408: I prefer to introduce a new type of array for fixed-length vectors so that we don't need to maintain the offset array.
and leave the BytesArray as it is...

skyzh: I'll leave it as a separate patch. `ArrayBuilder::with_capacity` only takes capacity, while "fixed-length vector" needs to supply the type information in advance (i.e., what are the cardinalities of the vectors). This requires an API change.

skyzh:
(ping for reviews, thanks)

wangrunji0408 (approved):
LGTM!

PR #867, opened 2025-01-08
feat(eval): support vector ops

skyzh:


src/binder/expr.rs
```diff
@@ -142,6 +142,12 @@ impl Binder {
             And => Node::And([l, r]),
             Or => Node::Or([l, r]),
             Xor => Node::Xor([l, r]),
+            Custom(name) => match name.as_str() {
+                "<->" => Node::VectorL2Distance([l, r]),
+                "<#>" => Node::VectorCosineDistance([l, r]),
+                "<*>" => Node::VectorDotProduct([l, r]),
```
wangrunji0408, line 148: Is there any standard for these operators?

It does not seem to be same as pgvector:
> Also supports inner product (<#>), cosine distance (<=>)

src/types/vector.rs
```diff
@@ -62,6 +62,40 @@ impl Deref for Vector {
...
+            .map(|(a, b)| (a.0 - b.0).powi(2))
+            .sum::<f64>();
+        F64::from(sum.sqrt())
+    }
+
+    pub fn cosine_distance(&self, other: &VectorRef) -> F64 {
+        let dot_product = self.dot_product(other);
+        let norm_self = self.norm();
+        let norm_other = other.norm();
+        F64::from(dot_product / (norm_self * norm_other))
```
wangrunji0408, line 85: Seems that `sqrt` can be optimized to once?

wangrunji0408 (commented):
LGTM so far.

src/binder/expr.rs
skyzh: ahh, I used AI autocompletion, so it might be wrong. will fix.

skyzh:
ready for review again :) also the preceding PR https://github.com/risinglightdb/risinglight/pull/866

PR #874, opened 2025-01-26
feat(planner): match vector indexes

skyzh:
This patch adds support for create index statements on vector columns and produces vector_index_scan plans. The executors are not implemented yet.


skyzh:
ready for review :)

skyzh:
but there's still difference between index scan and vector index scan: for vector index scans, we need to provide a constant vector; for normal index scan, it's just the order by column

skyzh:
let's use a single `IndexScan` for now and separate them when we add a normal index scan

wangrunji0408 (approved):
LGTM!

